PR #3, opened 2026-01-28
fix: unset COLORTERM to fix xterm.js rendering issues

eYdr1en:
## Summary
- Unsets `COLORTERM` in all PTY spawns to fix rendering issues in xterm.js

## Problem
When `COLORTERM=truecolor` is inherited from the outer terminal (e.g., Ghostty, iTerm2), Claude Code's Ink UI tries to use truecolor mode. However, xterm.js doesn't handle this correctly, causing rendering issues like gray/white backgrounds appearing where they shouldn't.

## Solution
Unset `COLORTERM` in the environment when spawning PTY processes so applications fall back to standard 256-color mode which xterm.js fully supports.

## Changes
- `screen-manager.ts`: Add `unset COLORTERM` to screen session environment
- `session.ts`: Add `COLORTERM: undefined` to all PTY spawn env objects

## Related
- https://github.com/xtermjs/xterm.js/issues/484

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/screen-manager.ts
+++ b/src/screen-manager.ts
@@ -258,7 +258,9 @@ export class ScreenManager extends EventEmitter {
 
     // Environment variables must be exported, not passed inline to nice
     // Using inline VAR=value before nice doesn't work correctly
+    // Unset COLORTERM to prevent truecolor issues - xterm.js doesn't handle inherited COLORTERM=truecolor well
     const envExports = [
+      'unset COLORTERM',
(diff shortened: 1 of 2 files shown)


Ark0N:
Fixed in 50b1718 — COLORTERM is now unset in all PTY spawns to prevent truecolor rendering issues in xterm.js.

PR #4, opened 2026-01-29
fix: add --verbose flag for stream-json output format

eYdr1en:
## Problem

The Ralph Loop wizard (plan orchestrator) was failing with the error:

```
When using --print, --output-format=stream-json requires --verbose
```

## Root Cause

Claude CLI v2.1.22+ changed its requirements - when using `--print` with `--output-format=stream-json`, the `--verbose` flag is now required.

## Solution

Added `--verbose` flag to the `runPrompt` args in `src/session.ts`:

```typescript
const args = [
  '-p',
  '--dangerously-skip-permissions',
  '--output-format', 'stream-json',
  '--verbose',  // Required for stream-json output format
];
```

## Testing

- Tested Ralph Loop wizard - now starts successfully without the error
- Verified with `claude --help` that --verbose is a valid flag

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -1213,6 +1213,7 @@ export class Session extends EventEmitter {
           '-p',
           '--dangerously-skip-permissions',
           '--output-format', 'stream-json',
+          '--verbose',  // Required for stream-json output format
         ];
         if (model) {
           args.push('--model', model);


Ark0N:
Fixed in 50b1718 — --verbose flag added to runPrompt() CLI args for Claude CLI v2.1.22+ compatibility.

PR #5, opened 2026-01-29
fix: ralph wizard agent files not found + JSON parse errors

eYdr1en:
### 1. Agent files not found for linked cases
When clicking on agent windows in the Ralph Loop wizard, users see "File not found" and "No result file found" errors for all agents.

**Root cause**: The ralph-wizard file endpoints only looked in `~/claudeman-cases/{name}`, ignoring **linked cases** (external folders linked via "Link Existing").

**Fix**: All three endpoints now check linked cases first:
- `/api/generate-plan-detailed` (for saving files)
- `/api/cases/:caseName/ralph-wizard/files`
- `/api/cases/:caseName/ralph-wizard/file/:filePath`

### 2. Execution optimizer agentType mismatch
SSE events were sent with `agentType: 'execution'` but files were saved to folder `execution-optimizer/`.

**Fix**: Consistent naming - SSE events now use `'execution-optimizer'`, frontend updated to match.

### 3. JSON parse "Bad control character" errors
```
[PlanOrchestrator] Execution optimizer JSON parse failed: Bad control character in string literal
```

**Root cause**: The `tryParseJSON` helper only escaped `\n`, `\r`, `\t` but Claude's responses can contain other control characters.

**Fix**: Enhanced `tryParseJSON` to escape ALL control characters (0x00-0x1F) using proper JSON escaping.

### 4. Prompts not visible while agents are running
Clicking on a running agent showed "File not found" because prompts were only saved on completion.

**Fix**: Prompts are now saved immediately when each agent **starts**, not when it completes. This allows viewing prompts while agents are still running.

## Files Changed
- `src/plan-orchestrator.ts` - agentType fix, JSON repair, save prompts on start
- `src/web/server.ts` - linked cases support for ralph-wizard endpoints
- `src/web/public/app.js` - frontend typeLabels/typeIcons update

## Testing
- TypeScript compiles without errors
- Tested with linked case - prompts visible while agents run
- Results saved correctly on completion

### DIFF
--- a/src/plan-orchestrator.ts
+++ b/src/plan-orchestrator.ts
@@ -364,18 +364,13 @@ export class PlanOrchestrator {
     this.screenManager = screenManager;
     this.workingDir = workingDir;
     this.outputDir = outputDir;
+    console.log(`[PlanOrchestrator] Constructor called with outputDir: ${outputDir || 'UNDEFINED'}`);
   }
 
   /**
-   * Save agent prompt and result to the output directory.
-   * Creates a folder for each agent with prompt.md and result.json files.
+   * Save agent prompt immediately when agent starts (so UI can show it while running).
    */
-  private saveAgentOutput(
-    agentType: string,
-    prompt: string,
-    result: unknown,
-    durationMs: number
-  ): void {
+  private saveAgentPrompt(agentType: string, prompt: string): void {
     if (!this.outputDir) return;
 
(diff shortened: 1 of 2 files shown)


Ark0N:
Fixed in 50b1718 — linked case resolution added to ralph-wizard endpoints, JSON repair for control characters.

PR #9, opened 2026-02-14
fix: resolve posix_spawnp and nested session spawn failures on macOS

k7cfo:
## Summary

Two issues prevent Claude from spawning in certain environments on macOS:

- **`posix_spawnp failed`**: `node-pty`'s `spawn-helper` prebuild binary loses its execute permission during `npm install`, causing every `pty.spawn()` call to fail. Fixed by adding `chmod +x` in `postinstall.js` for the darwin prebuilds.

- **`CLAUDECODE` env var leak**: When Claudeman is launched from within an existing Claude Code session (e.g. via Claude Code's Bash tool), the `CLAUDECODE=1` environment variable propagates into tmux and direct PTY child processes. Claude CLI detects this and refuses to start with _"cannot be launched inside another Claude Code session"_. Fixed by unsetting `CLAUDECODE` in both the tmux session command and the direct PTY `env` object.

## Changes

- `scripts/postinstall.js` — chmod `spawn-helper` prebuilds to `755` after install
- `src/tmux-manager.ts` — prepend `unset CLAUDECODE` to tmux session env exports
- `src/session.ts` — delete `CLAUDECODE` from env passed to `pty.spawn()` in all three spawn paths (tmux attach, direct PTY, shell mux attach)

## Test plan

- [x] Verified `npm run build` passes
- [x] Verified `claudeman web` starts and sessions spawn Claude successfully after fix
- [x] Tested on macOS arm64 (Darwin 25.2.0, Node 25.2.1)

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -6,9 +6,10 @@
  */
 
 import { execSync } from 'child_process';
-import { existsSync } from 'fs';
+import { chmodSync, existsSync } from 'fs';
 import { homedir, platform } from 'os';
-import { join } from 'path';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
 
 // ============================================================================
 // Configuration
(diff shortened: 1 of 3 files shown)


Ark0N:
Fixed in 50b1718 — chmod +x for spawn-helper in postinstall, CLAUDECODE env var unset in all PTY/tmux spawns.

PR #61, opened 2026-04-05
feat: file download support in browser file explorer

aakhter:
## Summary
- Add `?download=true` query param to existing file-raw endpoint for Content-Disposition attachment headers
- Add dedicated `GET /api/download` endpoint for filesystem file downloads
- Add download button to the file browser UI
- Fix 0-byte download bug by bypassing Fastify compression via `reply.raw` for file streams

## Context
When browsing files in the Codeman web UI, there was no way to download them to your local machine. This adds a download button to each file in the browser and the backend plumbing to serve files with proper Content-Disposition headers.

## Test plan
- [ ] Open file browser -- verify download button appears on files
- [ ] Click download -- verify file downloads with correct name and content
- [ ] Download a large file -- verify no 0-byte truncation
- [ ] Verify existing file-raw endpoint still works without `?download=true`

### DIFF
--- a/src/web/public/panels-ui.js
+++ b/src/web/public/panels-ui.js
@@ -2318,12 +2318,17 @@ Object.assign(CodemanApp.prototype, {
 
       const nameClass = isDir ? 'file-tree-name directory' : 'file-tree-name';
 
+      const downloadBtn = !isDir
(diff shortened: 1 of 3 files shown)


Ark0N:
Thanks! Cherry-picked the session-scoped download feature into master (f3cbe9b):

- Download button on file tree items (hover to reveal)
- `?download=true` query param on existing `/api/sessions/:id/file-raw` endpoint
- CSS styling for download button

**Not included:**
- Standalone `GET /api/download` endpoint — allows downloading arbitrary files from the filesystem with only a regex blocklist for protection. The session-scoped endpoint already validates paths are within the session's working directory, which is the right security boundary.

PR #71, opened 2026-04-24
fix(tmux-manager): use '|' separator in reconcileSessions

TeigenZhang:
## Summary
- `reconcileSessions()` invokes `tmux list-panes -a -F '#{session_name}\t#{pane_pid}'` and parses the result with `line.indexOf('\t')`.
- Under **non-tty execution contexts** (macOS LaunchAgent, systemd without `TTYPath`, Docker `exec` without TTY) tmux emits the `\t` in FORMAT strings as the literal two characters `\` + `t`, **not** as a tab (0x09). The parser therefore never matches, `activeSessions` stays empty, `reconcile` returns `{alive: [], dead: [], discovered: []}`, and `cleanupStaleSessions()` wipes every entry in `state.json` even though the underlying tmux sessions are still alive.
- Interactive `npm run dev` hides this because tmux's FORMAT parser resolves `\t` when stdout is a TTY.

## Root cause
Introduced in `88c415f` (perf: batch tmux reconciliation) — the regression only manifests in production deployments under launchd/systemd.

## Fix
Switch the separator to `|`. tmux passes it through verbatim in every environment, and `|` is rejected by tmux's own session-name validation so it cannot collide with the `codeman-<uuid>` / `claudeman-<uuid>` naming scheme.

## Test plan
- [x] Reproduce before fix on macOS LaunchAgent: service restart → `[StateStore] Cleaned up N stale session(s) from state` while `tmux ls` shows the sessions alive.
- [x] After fix: same setup → `[Server] Found N alive mux session(s) from previous run` → sessions restored with tokens and attached PTYs.
- [x] Interactive `npm run dev` still works: create session → kill-restart server → session restored.
- [x] `npm run typecheck` / `npm run lint` / `npm run build` all pass.

### DIFF
--- a/src/tmux-manager.ts
+++ b/src/tmux-manager.ts
@@ -904,14 +904,20 @@ export class TmuxManager extends EventEmitter implements TerminalMultiplexer {
     // Batch: single tmux call to get all session names + pane PIDs (replaces N per-session subprocess calls)
     const activeSessions = new Map<string, number>();
     try {
-      const output = execSync("tmux list-panes -a -F '#{session_name}\t#{pane_pid}' 2>/dev/null || true", {
+      // Use '|' rather than '\t' as FORMAT separator: under non-tty execution
+      // contexts (launchd on macOS, systemd without TTY) tmux emits '\t' in
+      // FORMAT strings as the literal two chars `\` + `t`, so `indexOf(tab-char)`
+      // never matches, `activeSessions` stays empty, and every known session is
+      // classified as dead — wiping state.json on restart. '|' is passed through
+      // verbatim in every environment and is not a valid tmux session-name char.
+      const output = execSync("tmux list-panes -a -F '#{session_name}|#{pane_pid}' 2>/dev/null || true", {
         encoding: 'utf-8',
         timeout: EXEC_TIMEOUT_MS,
       }).trim();
 
       for (const line of output.split('\n')) {
         if (!line) continue;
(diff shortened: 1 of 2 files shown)


Ark0N:
Thanks for the great catch and clean writeup, @TeigenZhang! The launchd / systemd-without-TTY repro and the choice of `|` (rejected by tmux's own session-name validation) both nailed it. State.json silently getting wiped on every production restart is exactly the kind of thing that's painful to track down.

I've pushed one follow-up commit (`bbb07be`) on top of yours since you had `maintainerCanModify` on:

- Extracted the inline parser into an exported `parsePaneList()` helper plus `PANE_LIST_SEP` / `PANE_LIST_FORMAT` constants in `src/tmux-manager.ts`. No behavior change in `reconcileSessions`.
- Added unit tests in `test/tmux-manager.test.ts` covering well-formed parsing, blank/missing-separator/non-numeric-pid skips, and — most importantly — a test that asserts a literal `\` + `t` in the input is **not** treated as a delimiter (the launchd regression scenario this PR fixes).

This locks in the contract so nobody can quietly "clean up" the separator back to `\t` later without CI catching it. `npm run typecheck` / `lint` / `format:check` all green, all 30 tests in `test/tmux-manager.test.ts` pass.

Hope that's OK with you — happy to drop the follow-up commit and merge as-is if you'd rather land the test in a separate PR. LGTM otherwise. 🚢

PR #74, opened 2026-04-24
refactor: pass envOverrides via tmux export instead of disk write

TeigenZhang:
## Summary
Before this change, `POST /api/sessions` persisted `envOverrides` (e.g. `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS`) to `<case>/.claude/settings.local.json` so Claude CLI would pick them up on spawn. That had two problems: UI dropdown state could drift from the file on disk, and the backend was silently writing into user project directories as a side-effect of opening a session.

This refactor makes envOverrides **ephemeral spawn-time state**:
```
UI dropdown → POST /api/sessions { envOverrides }
           → new Session({ envOverrides })
           → session._envOverrides
           → tmux-manager.buildEnvExports → `export KEY=<shellescape(VALUE)>` prepended to the pane's spawn command
```
- Survives `respawnPane` cycles via `session._envOverrides`
- Survives server restart via `SessionState.envOverrides` in `state.json`
- Introduces `session-ui.js#buildEnvOverrides(caseSettings, globalSettings)` as the single source of truth for the payload
- Removes `updateCaseEnvVars` import from `session-routes.ts` (now unused)

## Benefits
- No more stale `.claude/settings.local.json` files in user project directories
- UI state and spawn env are guaranteed consistent (one code path, one source)
- Frontend preview ("this is what will be exported") is trivially composable

## Test plan
- [x] Session spawn with `agentTeams` enabled → tmux pane sees `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` in env
- [x] Respawn pane preserves env overrides across restart
- [x] `~/.codeman/state.json` contains `envOverrides` for each session, restored on server restart
- [x] No writes to `<case>/.claude/settings.local.json` triggered by session creation

### DIFF
--- a/src/hooks-config.ts
+++ b/src/hooks-config.ts
@@ -85,6 +85,42 @@ export function generateHooksConfig(): { hooks: Record<string, unknown[]> } {
 }
 
 /**
+ * Remove a subset of env keys from .claude/settings.local.json.env if present.
+ * Used during the disk→tmux-setenv migration: when the caller is actively setting
+ * a fresh value for a Codeman-managed key, any stale disk entry for THAT KEY is
+ * superseded and should be removed. Keys NOT in `keysToRemove` are left alone
+ * (they may be user-managed). No-op if the file/keys don't exist.
+ */
+export async function stripCaseEnvKeys(casePath: string, keysToRemove: readonly string[]): Promise<void> {
+  if (keysToRemove.length === 0) return;
+
+  const settingsPath = join(casePath, '.claude', 'settings.local.json');
+  if (!existsSync(settingsPath)) return;
+
+  let existing: Record<string, unknown>;
+  try {
+    existing = JSON.parse(await readFile(settingsPath, 'utf-8'));
+  } catch {
+    return; // Malformed — don't rewrite it
+  }
+
+  const env = existing.env as Record<string, string> | undefined;
+  if (!env) return;
+
+  let changed = false;
+  for (const key of keysToRemove) {
+    if (key in env) {
+      delete env[key];
+      changed = true;
+    }
+  }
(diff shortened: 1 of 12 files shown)


Ark0N:
Thanks for this — the refactor is well-thought-out and addresses a real wart (silent writes into user case dirs). The `tmux setenv` approach is the right call: keeps secret values off the bash command line (`ps`-visible) while still surviving `respawn-pane`. The `__envOverrides` internal field with `CLAUDE_CODE_*`-only persistence filter is a sensible compromise between recovery and not leaking `OPENCODE_*` secrets into `state.json`. And the `buildEnvOverrides` helper is a nice cleanup of the four duplicated frontend call sites.

Two things I had to handle before merging:

1. **Merge conflict** with #73 (`thinkingEffort`) in `src/web/public/session-ui.js`. Resolved by keeping your `buildEnvOverrides()` helper — it already covers both `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` and `CLAUDE_CODE_EFFORT_LEVEL`, so the master-side inline block is fully replaced. Clean replacement, no behavioral drift.

2. **Test fix** in `test/session-manager.test.ts`: the `MockSession` class doesn't extend the real `Session`, so it didn't have `getEnvOverridesForPersist`. After `updateSessionState` started calling it, 19 tests failed with `TypeError`. Added a stub that returns `undefined` (matches the real method's contract for sessions without overrides).

Verified locally: typecheck, lint, format:check, build, and the four most-relevant test files (`session-manager`, `session-state`, `tmux-manager`, `tmux-restart-recovery`) all pass.

Merging now. 🙏

PR #75, opened 2026-04-24
feat(response-viewer): ASCII diagram wrap toggle, mobile code blocks, chrome-stripping fallback

TeigenZhang:
## Summary
Seven small commits that polish the Response Viewer (merged in #62) for daily use on mobile and desktop:

- **Card-style message separation + table wrap** (89a79a3) — messages get bordered cards; tables wrap in a horizontal-scroll container so they don't collapse on narrow viewports.
- **ANSI/CLI-chrome stripping fallback** (30b802b) — when the JSONL transcript isn't available, the viewer falls back to the terminal buffer; strip ANSI CSI/OSC/DCS, status bar, hints, spinner, and progress-bar glyphs so what the user sees is conversational text, not the TUI. Also captures the Claude CLI's real session ID for transcript lookup.
- **Mobile code-block wrap** (0bb4b85) — regular code blocks wrap on mobile by default (no more sideways scroll for every snippet); ASCII diagrams stay rigid with a horizontal scroll hint.
- **Per-block wrap toggle** (440b5c0) — add a toggle button on code blocks identified as ASCII diagrams so the user can flip between wrap (readable) and horizontal scroll (structure preserved).
- **Wrap-by-default + pinned toggle button** (dfe1dd9) — wrap is now the default; the toggle button lives outside the `<pre>` scroll container so it stays pinned to the visual right edge even while scrolling horizontally.
- **Narrow diagram detection** (e777ed9) — only box-drawing (`─-╿`) and block elements (`▀-▟`) trigger the diagram toggle. Previously arrows and geometric shapes in prose caused false positives.
- **Desktop eye icon** (49139d3) — the Response Viewer entry point now shows on desktop too (was mobile-only).

All changes are additive to `_renderMarkdown` / `_cleanTerminalBuffer` / `_preprocessAsciiArt` in `app.js` plus scoped CSS under `.rv-*` in `styles.css`. `_sanitizeHtml` (introduced in master) is still applied to markdown output.

## Test plan
- [x] Narrow mobile viewport: regular code blocks wrap, long responses scroll vertically only
- [x] ASCII diagram block shows wrap toggle; clicking flips wrap ↔ horizontal scroll
- [x] Table in response wraps in a scroll container instead of overflowing
- [x] When transcript file is missing, viewer shows clean conversation text (no ANSI, no TUI chrome)
- [x] Desktop browser shows the eye icon next to session tabs

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -1577,11 +1577,14 @@ export class Session extends EventEmitter {
         this._messages = this._messages.slice(-Math.floor(MAX_MESSAGES * 0.8));
       }
 
-      // Extract Claude session ID from messages (can be in any message type)
-      // Support both sessionId (camelCase) and session_id (snake_case)
+      // Extract Claude session ID from messages (can be in any message type).
+      // Support both sessionId (camelCase) and session_id (snake_case).
+      // The constructor seeds _claudeSessionId with this.id as a placeholder;
+      // once Claude CLI emits its real session ID, adopt it so JSONL lookups
+      // (e.g. /api/sessions/:id/last-response) can find the transcript file.
       const msgSessionId =
         ((msg as unknown as Record<string, unknown>).sessionId as string | undefined) ?? msg.session_id;
-      if (msgSessionId && !this._claudeSessionId) {
+      if (msgSessionId && msgSessionId !== this._claudeSessionId) {
         this._claudeSessionId = msgSessionId;
       }
 
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -999,24 +999,31 @@ class CodemanApp {
         // on mobile without collapsing into block-level cells.
         html = html.replace(/<table>/g, '<div class="rv-table-wrap"><table>')
                    .replace(/<\/table>/g, '</table></div>');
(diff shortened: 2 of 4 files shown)


Ark0N:
Thanks @TeigenZhang for this — the polish on mobile, the chrome-stripping fallback, and the wrap-toggle button are all great improvements.

Before merging I pushed one fix commit ([bdcd808](https://github.com/Ark0N/Codeman/pull/75/commits/bdcd808)) addressing two issues in `_renderMarkdown`:

1. **`src` is not defined** — the new code passes `src` to `_preprocessAsciiArt` but the parameter is `text`. The `try/catch` swallowed the `ReferenceError`, so `_renderMarkdown` always fell through to the plain-text `<pre>` branch. The new table-wrap, diagram-toggle, and ASCII preprocessing never actually ran in practice. ESLint can't catch this because `src/web/public/app.js` is in the eslint ignore list.
2. **XSS regression** — the description says `_sanitizeHtml` "is still applied", but the function was deleted in the refactor and the call removed, so `marked.parse()` output went straight into `innerHTML` unsanitized. Restored the helper and re-applied it before the table-wrap / diagram-toggle post-processing.

`npm run typecheck` and `npm run build` pass after the fix. Merging now.

PR #77, opened 2026-04-24
feat: improve Resume Conversation UX on mobile

TeigenZhang:
## Summary

On narrow screens the Resume Conversation list on the welcome page was hard to scan: the first prompt was clipped almost immediately by a single-line layout, and the working-dir subtitle was truncated at the end — exactly where the identifying project name lives. The `~/` shorthand also never fired on macOS because only `/home/` was handled.

This PR reworks the history item layout without changing the API:

- **Title** switches to a 2-line clamp so more of the first prompt is visible.
- **Subtitle** is now case-aware: workingDir is matched against loaded cases (incl. linked cases). Exact matches render as `#caseName`, subpaths as `#caseName/sub`, otherwise it falls back to the directory basename. `#` labels are styled in accent blue.
- **Home shorthand** now normalizes both `/home/<user>/` and `/Users/<user>/` to `~/` so macOS paths are collapsed.
- **Inline detail toggle**: each item gets a `⋯` button that expands an in-place panel with the full prompt, full path, timestamp, size, and short session id. Click again to collapse. Clicking the card body still triggers resume; the toggle uses `stopPropagation` to avoid accidental resume. The chevron rotates 90° in the expanded state.

Also fixes a small CSS bug where `.history-item-detail` used `display: flex` and therefore defeated the HTML `hidden` attribute's default `display: none` — an explicit `[hidden]` override was added.

## Test plan

- [x] Verify default render: title clamps to 2 lines, subtitle shows `#case` label when a case matches, basename otherwise.
- [x] Verify detail panel is collapsed by default.
- [x] Click `⋯` → panel expands with full prompt/path/meta. Click again → collapses.
- [x] Click card body (outside the button) → resume flow still runs.
- [x] macOS `/Users/<me>/...` paths display as `~/...` in the detail panel.
- [x] `npm run build` succeeds; assets rehash cleanly.

### DIFF
--- a/src/web/public/styles.css
+++ b/src/web/public/styles.css
@@ -2202,13 +2202,10 @@ body {
 
 .history-item {
   display: flex;
-  align-items: center;
-  gap: 0.75rem;
-  padding: 0.55rem 0.8rem;
+  flex-direction: column;
   background: rgba(255, 255, 255, 0.03);
   border: 1px solid rgba(255, 255, 255, 0.06);
   border-radius: 8px;
-  cursor: pointer;
   transition: all var(--transition-smooth);
   text-align: left;
 }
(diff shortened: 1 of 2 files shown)


Ark0N (approved):
Reviewed and verified locally on a mobile viewport (iPhone 13):

- Typecheck + lint clean; CI green
- Helper functions handle all edge cases: longest-prefix-wins for nested cases, basename fallback, null/empty safety, both Linux `/home/<user>/` and macOS `/Users/<user>/` normalized to `~/`
- Linked cases are picked up automatically since `/api/cases` merges them into the same `{name, path}` shape
- Live UI: detail panel hidden by default, expand toggles `aria-expanded`/`expanded` class/visibility cleanly, collapse works, no console errors
- The `[hidden]` CSS override is the correct fix for `display: flex` defeating the HTML `hidden` attribute

Nice work — the 2-line title clamp + `#caseName` subtitle + collapsible detail panel makes the welcome list much more readable on narrow screens. Thank you!

PR #80, opened 2026-05-11
fix: prevent tmux flicker on restart by matching existing window size

aakhter:
## Summary

When a PTY client re-attaches to an existing tmux session, the PTY size is hardcoded to **120x40**, so tmux resizes the window to match the smaller dimensions. The xterm.js client then immediately resizes back to its actual viewport on the next render tick — every restart causes a visible flicker and loses one repaint of scrollback content.

Two small changes fix this:

- **`src/session.ts`** — query the existing tmux window size via `tmux display -t <muxName> -p '#{window_width} #{window_height}'` before `pty.spawn`, fall back to 120x40 only if the call fails.
- **`src/tmux-manager.ts`** — drop the hardcoded `-x 120 -y 40` from `tmux new-session` so the initial size matches the first attaching client.

`execFileSync` is used (not `execSync`) so the muxName isn't shell-interpolated.

## Test plan

- [ ] Start a session, wait for buffer to fill (>40 rows of output).
- [ ] Hard-reload the page (Ctrl+Shift+R).
- [ ] Observe: no flicker, no buffer truncation.
- [ ] Restart with a non-default xterm size (e.g. wide window): on attach the PTY/tmux are sized to match.
- [ ] Kill tmux mid-attach (corner case): falls back to 120x40 gracefully (catch block in the `try { execFileSync … }`).

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -29,7 +29,7 @@
  */
 
 import { EventEmitter } from 'node:events';
-import { execSync } from 'node:child_process';
+import { execSync, execFileSync } from 'node:child_process';
 import { v4 as uuidv4 } from 'uuid';
 import * as pty from 'node-pty';
 import {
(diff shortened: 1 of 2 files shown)


Ark0N:
Thanks for the fix! Querying the existing tmux size before attach is exactly the right shape, and the `execFileSync` + bounded fallback is a nice touch. Merged.

PR #60, opened 2026-04-05
feat: keyboard UX improvements and session management

aakhter:
## Summary
- **Disable Ctrl+K kill-all-sessions**: Too easy to trigger accidentally (common browser shortcut), now requires confirmation or is disabled
- **Session rename with prefix preservation**: Rename sessions while keeping auto-generated prefixes intact
- **Shift+Enter / Ctrl+Enter multiline input**: Send multiline input via tmux `send-keys -H 0x0a` for reliable newline injection
- **Alt+1-9 session switching**: Standard keyboard shortcut to quickly switch between sessions, bypasses xterm.js key handling

## Test plan
- [ ] Verify Ctrl+K no longer kills all sessions without confirmation
- [ ] Rename a session -- verify prefix is preserved
- [ ] Type multiline input with Shift+Enter -- verify newlines are sent correctly
- [ ] Press Alt+3 -- verify it switches to the 3rd session tab

PR #81, opened 2026-05-12
fix(client): preserve inline rename input across tab re-renders

aakhter:
## Summary

Right-click → rename on a session tab injects an `<input>` into the tab name. While the user is typing, any incoming SSE event that triggers `renderSessionTabs()` (a sibling session updating, a hook firing, a status change) destroys the input element mid-keystroke and the user loses what they were typing.

This patch adds a `_inlineRenameActive` flag that:

- guards the two render paths (`renderSessionTabs()` and `_fullRenderSessionTabs()`) so they bail out early while a rename is in progress;
- is set true when the inline input mounts (`session-ui.js`);
- is cleared in `finishRename()`, which then explicitly calls `renderSessionTabs()` to restore the normal tab structure.

Also adds a re-entrance guard at the top of `finishRename()` so the blur event and the Enter keydown don't both fire it (was a latent double-call).

**Drive-by**: replaces `tabName.innerHTML = ''` with explicit child removal. The preceding `tabName.textContent = ''` already clears the element; this avoids an `innerHTML` write on a node that takes user-supplied content on the next line.

Follow-up to the inline-rename feature cherry-picked from #60.

## Test plan

- [ ] Open two sessions; in the second, generate continuous SSE traffic (start a build, run a long command).
- [ ] Right-click the first tab → rename, start typing.
- [ ] Verify the input is not destroyed by activity in the other session.
- [ ] Verify Enter and blur both finalize the rename correctly (only once — no double-call).
- [ ] Verify Esc still cancels (existing behavior, should be unaffected).

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -1844,6 +1844,8 @@ class CodemanApp {
   // ═══════════════════════════════════════════════════════════════
 
   renderSessionTabs() {
+    // Don't re-render while user is typing in the inline rename input
+    if (this._inlineRenameActive) return;
     this._debouncedCall('sessionTabs', this._renderSessionTabsImmediate);
   }
 
@@ -1988,6 +1990,7 @@ class CodemanApp {
   }
 
   _fullRenderSessionTabs() {
+    if (this._inlineRenameActive) return;
     const container = this.$('sessionTabs');
 
     // Clean up any orphaned dropdowns before re-rendering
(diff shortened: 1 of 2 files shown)


Ark0N:
Thanks for the careful fix! Verified that all SSE-triggered `renderSessionTabs()` call sites (in `panels-ui.js`, `subagent-windows.js`, `respawn-ui.js`) and the lone direct `_fullRenderSessionTabs()` in `settings-ui.js` are all covered by the guard. The re-entrance guard on `finishRename` is a nice catch on the latent async double-fire.

PR #82, opened 2026-05-12
feat: hostname-aware window title

aakhter:
## Summary

Set the browser tab title to `codeman:${hostname}` instead of the bare `Codeman` literal.
Useful for users running multiple Codeman instances across hosts (laptop, dev box, NAS) —
the OS hostname disambiguates which tab points at which backend, so you don't end up
typing into the wrong session.

Example:

| Before | After |
|--------|-------|
| `Codeman` | `codeman:laptop-aakhter` |
| `Codeman` | `codeman:nas02` |

## Implementation

- **`src/cli.ts`** — new `--title-hostname <hostname>` flag overrides the detected hostname (handy when `os.hostname()` returns something noisy, or when you just want a cosmetic name).
- **`src/web/server.ts`** — `WebServer` now accepts an optional `titleHostname` constructor arg (defaults to `os.hostname()`), composes `windowTitle = codeman:${titleHostname}`, and serves `/` and `/index.html` by templating that title into the cached `index.html` template. The `<title>` substitution is HTML-escaped via a small `escapeHtmlText()` helper so a hostname like `<script>` can't break out.
- **`src/web/public/notification-manager.js`** — title-flash logic now uses `this.originalTitle` instead of the hardcoded `Codeman` literal, so the tab flash respects the per-host title.
- **`scripts/browser-comparison.mjs` + `test/file-link-click.test.ts`** — expectations updated from `=== 'Codeman'` to a `startsWith('codeman:')` predicate so they pass regardless of host.

The index.html templating is intentionally narrow — it only substitutes the `<title>` tag and continues to serve everything else from the static template. No JS-side title injection, so the correct title shows from the very first paint and works without JavaScript.

## Test plan

- [ ] `npm run build && node dist/cli.js web -p 3000` — open in browser, tab title shows `codeman:<your-hostname>`.
- [ ] Same with `--title-hostname custom-name` — tab title shows `codeman:custom-name`.
- [ ] Trigger an idle/notification flash (background the tab while a session emits an event) — verify the flashing title alternates between `(N) codeman:<hostname>` and `codeman:<hostname>`, not the old `Codeman` literal.
- [ ] HTML escape: `--title-hostname '<script>'` → page title in the rendered HTML shows `codeman:&lt;script&gt;`, no DOM injection.

## Note on the test file diff size

`test/file-link-click.test.ts` shows ~49 prettier-reformat lines that aren't part of the feature — they're pre-existing prettier debt the local pre-commit hook required me to clear. The only behavioral change in that file is the `browserAvailable = ...startsWith('codeman:')` line.

### DIFF
--- a/scripts/browser-comparison.mjs
+++ b/scripts/browser-comparison.mjs
@@ -19,6 +19,10 @@ const PORTS = {
 
 const results = [];
 
+function isCodemanTitle(title) {
+  return typeof title === 'string' && title.startsWith('codeman:');
+}
+
 function logSection(title) {
   console.log('\n' + '='.repeat(60));
   console.log(`  ${title}`);
@@ -88,7 +92,7 @@ async function main() {
       const page = await playwrightBrowser.newPage();
       await page.goto(`http://localhost:${PORTS.playwright}`);
       const title = await page.title();
-      if (title !== 'Codeman') throw new Error(`Expected Codeman, got ${title}`);
+      if (!isCodemanTitle(title)) throw new Error(`Expected codeman:<hostname>, got ${title}`);
       await page.close();
     });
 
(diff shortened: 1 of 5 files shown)


Ark0N:
Thanks — this is genuinely useful for me too (I bounce between several Codeman hosts). The implementation is exactly the right shape: server-side HTML templating so the correct title paints on first byte (no JS-side flicker), narrow HTML-escape scoped to title element content, and the notification-manager change so the flashing title also respects the per-host name. `--title-hostname` override is a nice escape hatch for noisy `os.hostname()` outputs.

PR #92, opened 2026-05-19
fix: decode project keys with longest-match backtracking

TeigenZhang:
## Summary

When two sibling directories share a prefix (e.g. `diary/` and `diary-app/`), the greedy shortest-match decoder in `decodeProjectKey()` picks the shorter name and then fails to resolve the remainder. As a result:

- The homepage **Resume Conversation** list shows those `workingDir`s as `$HOME`
- Resume targets the wrong folder

## Fix

Switch to recursive backtracking with **longest-join-first** at each segment boundary; require every step to be a real directory on disk. Keep the greedy path as a fallback for deleted directories (so history of deleted projects still resolves).

## Test plan

- [x] `tsc --noEmit` passes
- [x] `npm run lint` passes
- [x] Manual verification: sibling directories `~/foo/diary/` and `~/foo/diary-app/` both resolve correctly in the Resume Conversation list and Resume button targets the right folder.

### DIFF
--- a/src/web/routes/session-routes.ts
+++ b/src/web/routes/session-routes.ts
@@ -1324,42 +1324,76 @@ export function registerSessionRoutes(
    * Claude CLI encodes both '/' and '_' as '-', so each '-' in the key could be
    * any of: '/' (path separator), '_' (underscore), or '-' (literal dash).
    *
-   * Strategy: look-ahead matching. At each '-', try consuming multiple segments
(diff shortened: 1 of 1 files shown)


Ark0N:
Thanks for the thoughtful fix! Good catch on the sibling-prefix collision — the longest-match backtracking with the greedy fallback for deleted dirs is exactly the right shape. Merging.

PR #93, opened 2026-05-19
fix: drop phantom ended-tab stubs, trust server as source of truth

TeigenZhang:
## Summary

The client cached open session tabs in `localStorage` and resurrected any that the server no longer knew about as grayed-out "ended" stubs. On **multi-device use** (close tab on mobile, open desktop) this left stale phantom tabs the user had to manually dismiss every time.

## Fix

Drop the client-side resurrection path entirely; let the server be the source of truth for which sessions are open.

Removed:
- `_restoreEndedTabs` / `_saveTabMetadata`
- The `session._ended` branch in `selectSession`
- The `data-ended` render attribute and its matching CSS rule

Also clears the legacy `localStorage` key on init to purge stale entries left by previous client versions.

Net: **-44 / +6 lines** — removes complexity while fixing the UX bug.

## Test plan

- [x] `tsc --noEmit` passes
- [x] `npm run lint` passes
- [x] Manual verification: close a tab on mobile, refresh desktop — no phantom grayed-out tab appears.
- [x] Manual verification: existing legacy `localStorage` entries from older clients are cleared on next page load.

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -1794,8 +1794,10 @@ class CodemanApp {
       }
     });
 
-    // Restore tabs that were open before refresh but are no longer on the server
-    this._restoreEndedTabs();
+    // Server is source of truth for open sessions — don't resurrect stale tabs
+    // from localStorage (would show phantom "ended" tabs when a session was closed
+    // on another device).
+    try { localStorage.removeItem('codeman-tab-meta'); } catch {}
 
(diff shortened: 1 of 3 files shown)


Ark0N:
Great input — thank you! Clean removal: the localStorage resurrection was the only thing that set `_ended: true` on the client, so dropping it along with all the consumers is consistent. Server-side ended status (via `.tab-status.ended`) is preserved for sessions the server still knows about. Merging.

PR #94, opened 2026-05-19
feat: View all in folder modal for Resume Conversation

TeigenZhang:
## Summary

The homepage Resume Conversation list dedups history at **3 sessions per project** so the top of the page doesn't get dominated by a single hot folder. But once you want to dig into a specific project's older conversations, there is no drill-down — they're hidden.

This adds a **"View all in this folder"** action that opens a modal listing the complete history for a single project, paginated.

### Backend (`src/web/routes/session-routes.ts`)

`GET /api/history/sessions` now accepts:
- `projectKey` — restrict to a single project
- `offset` / `limit` — pagination

Single-folder mode bypasses the 50-cap and returns `{ sessions, total }`.

`projectKey` is validated against `^[A-Za-z0-9_-]+$` before any filesystem lookup, so the parameter cannot be used for path traversal.

### Frontend (`src/web/public/terminal-ui.js`, `styles.css`)

- The history-item detail panel gains a **"View all in this folder"** button when `projectKey` is present.
- Clicking it opens a modal listing 20 sessions per page with a **"Show more"** button for pagination.
- Modal items reuse `_buildHistoryItem` with `showViewAll: false` to avoid recursive entry points.

## Test plan

- [x] `tsc --noEmit` passes
- [x] `npm run lint` passes
- [x] Manual: open Resume Conversation, expand a project with >3 sessions, click "View all in this folder", verify modal lists more than 3 sessions and pagination works.
- [x] Manual: confirm modal items do NOT render their own "View all" button (no recursion).
- [x] Security: requests with `projectKey=../../etc` are rejected by the regex before reaching the filesystem.

### DIFF
--- a/src/web/public/styles.css
+++ b/src/web/public/styles.css
@@ -2370,6 +2370,56 @@ body {
   color: var(--text);
 }
 
+.history-detail-actions {
+  margin-top: 0.5rem;
+}
+
+.history-view-all-btn {
+  width: 100%;
+  padding: 0.45rem 0.75rem;
+  background: rgba(99, 179, 237, 0.08);
+  border: 1px solid rgba(99, 179, 237, 0.25);
+  border-radius: 6px;
+  color: rgba(99, 179, 237, 0.95);
+  font-size: 0.78rem;
+  font-weight: 500;
+  cursor: pointer;
+  transition: background var(--transition-smooth), border-color var(--transition-smooth);
+}
+
+.history-view-all-btn:hover {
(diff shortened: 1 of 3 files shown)


Ark0N:
Great input — thank you! Nice refactor: extracting `scanProjectDir` keeps the global and drill-down code paths in sync, and the `projectKey` regex closes path traversal before the filesystem call. The recursion guard via `showViewAll: false` is a clean way to reuse `_buildHistoryItem` inside the modal. Merging.

PR #96, opened 2026-05-25
fix(tmux): isolate sessions on a dedicated socket + raise pane nofile limit (fixes new-session crash after tmux upgrade)

TeigenZhang:
## Symptom

After upgrading tmux to the latest version, **opening a new conversation immediately errored out** — the freshly spawned CLI process exited right away, so no session could start.

## Root cause

Under macOS launchd, the newer tmux hands panes a low soft `nofile` limit (256). Recent Claude Code exits immediately when it inherits such a low file-descriptor limit, so every newly spawned session died on startup.

Separately, Codeman's sessions lived on the **default** tmux server intermixed with the user's own tmux sessions, and carried a per-session `tmuxSocket` tag that could drift from physical reality — causing live sessions to be wrongly marked dead ("tab shows no session found") and duplicate "Restored:" tabs.

## Fix

1. **Raise the pane `nofile` limit** (`ulimit -Sn`) before launching the CLI, so the new tmux + launchd combo no longer crashes Claude Code on spawn.
2. **Isolate all Codeman sessions onto a dedicated socket** (`tmux -L codeman`), fully separated from the user's default tmux server.
3. **Unify to a single socket** — remove the drift-prone per-session `tmuxSocket` field. `reconcileSessions()` collapses from a multi-socket scan to a single `list-panes` query; `loadSessions()` strips the obsolete field. Socket name exposed as `TmuxManager.muxSocket`, overridable via `CODEMAN_TMUX_SOCKET`.
4. Route two sibling bare-`tmux` call sites through the socket (same regression class): `queryTmuxWindowSize()` (was silently falling back to 120×40 on re-attach, losing scrollback) and the `send-key` route (Shift+Enter / Ctrl+Enter newline).
5. SSH chooser scripts (`tmux-manager.sh`, `tmux-chooser.sh`) route every tmux call through `tmux -L $CODEMAN_TMUX_SOCKET`.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm test -- test/tmux-manager.test.ts test/tmux-window-size-query.test.ts` → 46 passing
- [x] Verified on a live deployment: new sessions spawn without the nofile crash; existing tabs reconcile on the dedicated socket; no "session not found" or duplicate "Restored:" tabs.

### DIFF
--- a/scripts/tmux-chooser.sh
+++ b/scripts/tmux-chooser.sh
@@ -32,6 +32,13 @@ set -e
 CODEMAN_STATE="$HOME/.codeman/state.json"
 CODEMAN_SESSIONS="$HOME/.codeman/mux-sessions.json"
 
+# Dedicated tmux socket all Codeman sessions live on. MUST match
+# DEFAULT_CODEMAN_TMUX_SOCKET / CODEMAN_TMUX_SOCKET in src/tmux-manager.ts —
+# otherwise list-sessions would enumerate the user's default tmux server
+# (missing the real Codeman sessions, surfacing unrelated ones).
+CODEMAN_TMUX_SOCKET="${CODEMAN_TMUX_SOCKET:-codeman}"
+TMUX_CMD=(tmux -L "$CODEMAN_TMUX_SOCKET")
+
 
 # iPhone 17 Pro portrait width (conservative)
 MAX_WIDTH=44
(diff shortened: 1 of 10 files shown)


Ark0N:
Thank you very much for this — it's an excellent, thorough fix. 🙏

I reviewed the full diff and it's in great shape:
- **nofile fix** (`buildNofileLimitCommand`) is prepended to both the spawn and respawn command strings — correct coverage of both paths.
- **Single-socket isolation** is threaded consistently through every `tmux` call site (manager, `queryTmuxWindowSize`, the send-key route, and both shell chooser scripts), with the socket name validated (`SAFE_TMUX_SOCKET_PATTERN`) and shell-escaped — injection-safe.
- Dropping the per-session `tmuxSocket` field plus the `loadSessions()` dedup-by-`muxName` pass cleanly retires the drift that caused the phantom "Restored:" twins.
- Tests are great — the hex-allowlist negative case and the `-L <socket>`-must-lead call-shape assertions are exactly the right regression guards. CI is green.

One tiny non-blocking nit for a future pass: the `buildNofileLimitCommand` JSDoc block is in Chinese while the rest of the file is English — worth translating for consistency.

Merging now. Thanks again for the careful work and detailed write-up!

PR #98, opened 2026-05-25
feat(web): one-click copy button on response-viewer code blocks

TeigenZhang:
Adds a one-click **copy** button to every fenced code block in the response viewer.

- Each code block is wrapped in a positioned `.rv-code-wrap` toolbar pinned top-right, **outside** the `<pre>` scroll container so the buttons stay put during horizontal scroll. All blocks get a copy button; ASCII diagrams keep their existing line-wrap toggle alongside it.
- `_copyText()` prefers the async Clipboard API and falls back to a hidden-`textarea` + `execCommand` path, so copy works over plain HTTP too.
- The button shows a 1.5s ✓ / ✕ feedback state after each attempt.

## Test plan

- [x] esbuild parses `app.js` / `styles.css` cleanly
- [x] Verified in the UI: copy button appears on code blocks, copies the raw source, shows ✓ on success (both HTTPS and plain HTTP)

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -1128,22 +1128,40 @@ class CodemanApp {
         const DIAGRAM_CHAR = /[─-╿▀-▟]/;
         const tmpl = document.createElement('template');
         tmpl.innerHTML = html;
+        // Every fenced code block gets a positioned wrapper with an action
+        // toolbar pinned to its top-right corner. The toolbar lives OUTSIDE the
+        // <pre> scroll container so its buttons stay put during horizontal
+        // scroll. All blocks get a one-click copy button; ASCII diagrams keep
+        // the additional line-wrap toggle.
         tmpl.content.querySelectorAll('pre > code').forEach((code) => {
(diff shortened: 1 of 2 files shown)


Ark0N:
Thank you very much for this — really clean, well-scoped contribution! 🙏

I reviewed it end to end and everything checks out:
- The DOM restructure preserves the existing diagram line-wrap toggle: diagrams keep the `rv-diagram-wrap` class, so the delegated handler's `closest('.rv-diagram-wrap')` / `querySelector('pre.rv-diagram')` still resolve.
- CSS specificity is correct — the new `.rv-code-wrap.rv-diagram-wrap > pre.rv-diagram` rule properly widens the padding reserve to `76px` for the two-button case, and `.rv-code-actions .rv-wrap-toggle { position: static }` neatly re-flows the toggle into the flex toolbar.
- The `execCommand` fallback in `_copyText()` is a nice touch — copy works over plain HTTP (tunnel) too.
- Reusing the `⎘` glyph keeps it visually consistent with the file-preview copy button.

CI is green (Typecheck & Lint). Merging now. Thanks again! 🎉

PR #99, opened 2026-05-25
fix(web): stop auto-sending Ctrl+L from session selection paths

TeigenZhang:
## Symptom

Quickly refreshing the page (or an SSE reconnect firing in quick succession) would sometimes **silently wipe the current conversation** — the active Claude session got cleared with no warning or confirmation.

## Root cause

Claude Code 2.x treats Ctrl+L (`\x0c`) as a two-step "clear conversation" command (first press shows a confirmation prompt, second press clears). The frontend fired `\x0c` from three places to force Ink to redraw stale CUP-positioned frames in the tailed buffer. When a fast page refresh / reconnect ran the same path **twice within Claude's confirmation window**, the first `\x0c` opened the clear prompt and the second `\x0c` confirmed it — destroying the conversation.

## Fix

Removed the `\x0c` sends from:
- `selectSession()` — the main offender, runs on every tab switch & page reload
- `restoreTerminalSize()` — manual "restore size" button
- `sendPendingCtrlL()` — dead code path (`pendingCtrlL` was never populated)

**Trade-off:** occasional stale Ink frame immediately after refresh; the user's first keypress causes Ink to redraw and the artifact vanishes. Losing the conversation silently is far worse than a brief cosmetic glitch.

## Test plan

- [x] esbuild parses `app.js` / `terminal-ui.js` cleanly
- [x] Verified: tab switches / page reloads no longer risk clearing the conversation; transient redraw artifact self-heals on first keypress

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -2643,19 +2643,15 @@ class CodemanApp {
         });
       }
 
-      // Fire-and-forget resize + Ctrl+L to force Ink redraw.
-      // Tailed buffers accumulate stale CUP-positioned Ink frames that overlap
-      // in the viewport (e.g. duplicate "bypass permissions" bars). Ctrl+L
-      // triggers a full Ink redraw which overwrites all stale frame content.
-      // sendResize may be a no-op if dimensions match, so Ctrl+L is essential.
-      this.sendResize(sessionId).then(() => {
-        if (selectGen !== this._selectGeneration) return;
-        fetch(`/api/sessions/${sessionId}/input`, {
-          method: 'POST',
-          headers: { 'Content-Type': 'application/json' },
-          body: JSON.stringify({ input: '\x0c' })
-        }).catch(() => {});
-      });
+      // Fire-and-forget resize to nudge Ink via SIGWINCH on real size changes.
+      // Previously we also sent Ctrl+L (\x0c) here to force a full Ink redraw,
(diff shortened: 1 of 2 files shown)


Ark0N:
Thank you very much — this is an excellent catch and a really well-reasoned fix. 🙏

Silently wiping the conversation on a fast refresh / SSE reconnect is about the worst failure mode we could have, so trading it for a transient stale Ink frame that self-heals on the next keypress is clearly the right call.

I verified the whole change before merging:
- **`pendingCtrlL` really is dead** — it's only ever *read* inside `sendPendingCtrlL`, never assigned anywhere, so the `if (!this.pendingCtrlL)` guard always returned early. The function was already a no-op in practice and has zero callers, so the explicit stub changes nothing.
- The **user-initiated Ctrl+L** keymap in `input-cjk.js` is correctly left intact — only the automatic/programmatic sends are gone.
- No correctness regression in the anti-flicker pipeline: redraw correctness there comes from resize/SIGWINCH + the buffering layers; the `selectSession` `\x0c` was purely a cosmetic force-redraw.
- The backend `CTRL_L_PATTERN` filter only strips form-feed from PTY *output*, so it's unrelated to these input sends.

Love the "Do NOT re-introduce Ctrl+L here" guard comments too — that'll save a future regression. CI is green. Merging now, thanks again! 🎉

PR #100, opened 2026-05-31
fix: avoid event-loop stalls from synchronous tmux/ps calls

TeigenZhang:
## Summary

The session manager called `execSync` for tmux mouse-mode and `list-panes` queries on the main thread. Under load (many sessions / agent panes) these synchronous spawns block Node's event loop, stalling SSE broadcasts and PTY I/O.

This converts those calls to async (`execAsync`) and adds a lightweight `event-loop-monitor` utility that samples loop-delay and logs when a stall threshold is exceeded, making future regressions observable.

## Changes
- `tmux-manager.ts`: `enableMouseMode`/`disableMouseMode`/`listPanes` now use `execAsync` instead of `execSync`
- `utils/event-loop-monitor.ts` (new): periodic event-loop delay sampler
- `utils/index.ts`: export the monitor
- `server.ts`: start the monitor on boot

## Test plan
- [x] `tsc --noEmit` passes
- [ ] Observe no main-thread stalls under multi-session load

### DIFF
--- a/src/tmux-manager.ts
+++ b/src/tmux-manager.ts
@@ -1153,19 +1153,23 @@ export class TmuxManager extends EventEmitter implements TerminalMultiplexer {
     }
 
     try {
-      const psOutput = execSync(`ps -o rss=,pcpu= -p ${session.pid} 2>/dev/null || echo "0 0"`, {
-        encoding: 'utf-8',
-        timeout: EXEC_TIMEOUT_MS,
-      }).trim();
+      const psOutput = (
+        await execAsync(`ps -o rss=,pcpu= -p ${session.pid} 2>/dev/null || echo "0 0"`, {
+          encoding: 'utf-8',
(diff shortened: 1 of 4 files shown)


Ark0N:
Thank you as always for contributing, @TeigenZhang! 🙏

Reviewed and verified locally on the PR branch:
- ✅ `tsc --noEmit` clean
- ✅ `npm run lint` clean
- ✅ `npm run format:check` clean
- ✅ `npm run build` (esbuild) succeeds
- ✅ All internal callers of `listPanes`/`enableMouseMode`/`disableMouseMode`/`syncMouseMode` updated to `await`; no external callers left unhandled

Great catch on the blocking `execSync` calls stalling the event loop under multi-session load — and the `event-loop-monitor` is a nice touch for making future regressions observable. Merging now.

PR #102, opened 2026-05-31
feat(web): response-viewer transcript fallback + code-block rendering

TeigenZhang:
## Summary

Improves the response viewer when the JSONL transcript is unavailable (it has to fall back to the raw terminal buffer) and hardens code/diagram rendering.

## Changes
- `app.js` `_cleanTerminalBuffer()`: strip ANSI escapes + Claude CLI chrome (status bar, spinner, progress bar, prompt glyphs) so the fallback renders clean conversational text
- `app.js` `_preprocessAsciiArt()`: wrap box-drawing/block-element diagrams in fenced code blocks (narrow trigger excludes arrows/geometric shapes common in prose) so marked.js preserves whitespace
- `styles.css`: extend `.rv-text` rules to `.response-viewer-body` so fallback-rendered content gets the same typography, code-block, and table styling

## Notes
Built on top of the existing response-viewer (copy button / wrap toggle already upstream); this fills the transcript-missing gap.

## Test plan
- [x] `tsc --noEmit` passes (no .ts touched)
- [ ] Open a response with no JSONL transcript, confirm clean render
- [ ] Render a box-drawing diagram, confirm whitespace preserved


Ark0N:
Thank you as always for contributing, @TeigenZhang! 🙏

The response-viewer CSS overhaul is lovely — proportional prose font with monospace kept for code, the heading/blockquote/code styling, the readable max-width, and the smoother slide-in animation all make the viewer much nicer to read. The `_renderMarkdown` null-safety fix (`text → src = text || ''`) is a good catch too.

One heads-up from review: this branch forked before realizing that `_cleanTerminalBuffer` and `_preprocessAsciiArt` already landed on master back in #75, so it had re-added byte-identical copies of both. In a JS class body the later definition wins, so the duplicates were inert dead code. I pushed a small cleanup commit (`bdeab49`) removing them, leaving just the genuinely-new CSS + null-safety work.

Verified after cleanup: each method now defined exactly once, `tsc`/`lint`/`format`/`build` all clean, `node --check` clean, and a test-merge confirms #100/#101 are preserved. Merging now.

PR #114, opened 2026-06-10
feat(codex): add Codex (OpenAI CLI) run-mode foundation

aakhter:
Adds **Codex (OpenAI CLI)** as a first-class session mode alongside Claude / Shell / OpenCode, mirroring the existing OpenCode run-mode integration.

### What's included
- **`codex-cli-resolver`** — locates the `codex` binary across common install dirs and augments PATH (mirrors `opencode-cli-resolver`).
- **Types** — `SessionMode` gains `'codex'`; new `CodexConfig` (`model`, `resumeSessionId`, `dangerouslyBypassApprovals`, `renderMode`), persisted in `SessionState` and threaded through `CreateSessionOptions` / `RespawnPaneOptions`.
- **Validation** — `CodexConfigSchema`, `CODEX_` added to the env-var prefix allowlist, `codex` in the create/quick-start mode enums, and a `codexDangerouslyBypassApprovals` setting.
- **tmux launch** — `buildCodexCommand`, `setenv` for `OPENAI_API_KEY`/`CODEX_*` (keeps secrets out of `ps`/history), truecolor `COLORTERM`, and Codex PATH resolution.
- **Session + routes** — availability check with a clear install hint, config passthrough, a tmux-required guard, and Codex is correctly skipped by Claude-only parsers (Ralph / respawn / token parsing).
- **UI** — a "Run CX" run-mode option and a dedicated **Codex CLI** settings tab with the bypass-approvals toggle; `GET /api/codex/status`.

### Scope
Foundation only — Codex-specific terminal redraw handling and xterm snapshot/replay are intentionally **out of scope** and tracked separately, to keep this PR focused and reviewable. Gemini is not included.

### Verification
- `tsc --noEmit`, `eslint`, `prettier --check`, `check:frontend-syntax` — all clean
- `test:ci` (full unit/integration suite) — **2712 passed, 0 failed**, incl. new `test/run-mode-ui.test.ts` (vm-based, covers run-mode selection + Codex quick-start wiring)
- Server boot smoke (`/api/status`) — OK

### DIFF
--- a/src/mux-interface.ts
+++ b/src/mux-interface.ts
@@ -14,6 +14,7 @@ import type {
   ClaudeMode,
   SessionMode,
   OpenCodeConfig,
+  CodexConfig,
   EffortLevel,
 } from './types.js';
 
@@ -62,6 +63,7 @@ export interface CreateSessionOptions {
   claudeMode?: ClaudeMode;
   allowedTools?: string;
   openCodeConfig?: OpenCodeConfig;
+  codexConfig?: CodexConfig;
   /** When restoring after reboot, resume a previous Claude conversation by its session ID */
   resumeSessionId?: string;
   /** Extra env vars exported before launching the CLI (e.g., CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS). Ephemeral — not written to disk. */
@@ -80,6 +82,7 @@ export interface RespawnPaneOptions {
   claudeMode?: ClaudeMode;
   allowedTools?: string;
   openCodeConfig?: OpenCodeConfig;
+  codexConfig?: CodexConfig;
   /** Resume a previous Claude conversation when respawning */
   resumeSessionId?: string;
   /** Extra env vars exported before launching the CLI (preserved across respawns). */
--- a/src/session.ts
+++ b/src/session.ts
@@ -125,7 +125,7 @@ const CTRL_L_PATTERN = /\x0c/g;
 const NEWLINE_SPLIT_PATTERN = /\r?\n/;
 
 /** True for external-CLI run modes (non-Claude) that use their own TUI and output format. */
-function isExternalCliMode(mode: SessionMode): boolean {
+export function isExternalCliMode(mode: SessionMode): boolean {
   return mode === 'opencode' || mode === 'codex';
 }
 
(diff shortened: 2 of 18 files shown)


Ark0N:
Reviewed in depth — thanks @aakhter, another excellent contribution! 🙏 The integration faithfully mirrors the OpenCode run-mode pattern everywhere it matters: the resolver, tmux `setenv` secret injection (keys out of `ps`/history), schema-validated `codexConfig` with shell-safe regexes, the `CODEX_` env-prefix allowlist gating, the tmux-required guard, and skipping Claude-only parsers via `isExternalCliMode()`. The `runMode` accessor fix is a nice catch too — the old object-literal getter was silently flattened to a static value by `Object.assign`, which is exactly the server-sync desync your new test pins down.

One functional blocker plus a few parity gaps, fixed in 8c7a978 on this branch:

- **Blocker — response envelope:** `runCodex()` read raw shapes (`status.available`, `data.sessionId`), but the global `preSerialization` hook wraps every payload in `{ success, data }` — so the UI printed "Codex CLI not found" unconditionally and could never start a session. Fixed to match `runOpenCode` (`.data.available` / `data.data.sessionId`), and the test mocks now use the real wire shape (plus a `selectSession` assertion) so envelope drift fails the test going forward.
- **Guard parity:** the ralph-config guard, all three respawn guards, and six restore/setup guards in `server.ts` only excluded `'opencode'` — a codex session could get a Ralph tracker or respawn controller attached, and output-silence idle detection would misfire on a quiet codex TUI. All now use the exported `isExternalCliMode()`.
- **UI parity:** `cx` tab badge, "Kill Tmux & Codex" close-dialog title, and the missing CSS (`.run-mode-dot.codex` was invisible; `.mode-codex` button colors, `.tab-mode.codex`) — codex gets purple.

Verified end-to-end on an isolated instance with a stub `codex` binary: 10/10 Playwright checks (menu option + dot, "Run CX" label/styling, session created **and auto-selected**, `cx` badge, TUI output streaming, ralph + respawn guards rejecting codex), and `--dangerously-bypass-approvals-and-sandbox --model gpt-5.2` observed on the spawned command line. `tsc`, lint, prettier, frontend-syntax, and the CI suite are green. Merging!

PR #116, opened 2026-06-10
fix(codex): strip alt-screen + scrollback-erase from the codex byte stream

aakhter:
## Problem

When running a Codex session, the conversation history disappears (and becomes unscrollable) on every tab switch / pane refresh.

Codex's TUI emits, during startup and on each repaint:
- alternate-screen toggles — DECSET/DECRST `47` / `1047` / `1049`
- scrollback-erase — `CSI 3 J`
- mouse-tracking enables — `?1000`–`1007`

xterm.js faithfully obeys these: it switches to the **alternate** buffer (which has no scrollback), wipes the saved lines, and forwards the scroll wheel to the codex process. The net effect is that the user's prior conversation/shell output both vanishes and — even when still present in the buffer — can't be scrolled back to.

## Fix

Strip just those sequences from the codex byte stream in two places, leaving the **visible-viewport** erases (`2J` / `J`) intact so codex can still repaint its own rows:

- **`Session._handleTerminalOutput`** — filters the live SSE/WS stream and the persisted terminal buffer at the source, for `mode === 'codex'`.
- **`GET /api/sessions/:id/terminal`** — applies the same strip (`ALT_SCREEN_TOGGLE_PATTERN` / `ERASE_SCROLLBACK_PATTERN`) to the replayed buffer, so a tab-switch replay keeps full scrollback. Mirrors the live-stream strip.

Only `mode === 'codex'` is affected; Claude/OpenCode/shell paths are untouched.

## Tests

Adds `test/codex-terminal-output.test.ts`:
- alt-screen toggles and `3J` are removed; `2J` / `J` (visible-viewport erases) are preserved
- a `?1049h … 2J … 3J … ?1049l` redraw collapses to just the visible content
- Ctrl+L redraws and native codex prompt/status/resume-picker redraws pass through without row-repair mangling

`npm run test:ci` passes locally (2735 passed, 0 failures); `tsc --noEmit`, `eslint`, `prettier --check`, and the frontend-syntax check are all clean.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -1052,6 +1052,32 @@ export class Session extends EventEmitter {
   }
 
   private _handleTerminalOutput(data: string): void {
+    // Codex emits sequences that wipe xterm.js scrollback, plus mouse-tracking
+    // enables that hijack the scroll wheel so the user can't reach scrollback:
+    //   - \x1b[?1049h / \x1b[?47h / \x1b[?1047h: switch to the alt buffer (no
+    //     scrollback) — \x1b[?...l switches back.
+    //   - \x1b[3J: erase saved lines (scrollback). (\x1b[2J / \x1b[J — erase
+    //     the visible viewport — are left intact; the TUI repaints those rows.)
+    //   - \x1b[?1000h / 1002h / 1003h / 1005h / 1006h / 1007h: mouse-tracking
+    //     modes (X10, button-event, any-event, UTF-8, SGR, alt-scroll). Once on,
+    //     xterm.js forwards wheel events to codex instead of scrolling the
+    //     viewport, so the conversation is in scrollback but unreachable.
+    //     (Focus events at ?1004 are left alone — codeman uses them for
+    //     active-tab detection.)
+    // Strip them at the source so neither the persisted buffer nor the live
+    // SSE/WS stream carries them, keeping everything in the main buffer with
+    // scrollback intact. Codex's cursor-positioned redraws overwrite only the
+    // cells they actually target, so the non-erased rows keep their content.
+    if (this.mode === 'codex') {
+      data = data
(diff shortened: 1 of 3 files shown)


Ark0N (commented):
Reviewed. The approach is right — strip at the source so neither the persisted buffer nor the live stream carries the scrollback-killing sequences, with the replay strip as backstop — and the 13 captured-stream tests are genuinely good coverage. Two gaps, both fixed in `7eda39e` pushed to this branch:

**1. Sequences split across PTY chunk boundaries slipped past the live strip.** `_handleTerminalOutput` filters per chunk, but chunk boundaries are arbitrary: `\x1b[?1049h` can arrive as `\x1b[?104` + `9h`. xterm's parser is streaming, so it reassembles and obeys the leaked toggle — and since the matching `?1049l` (arriving intact) *does* get stripped, the terminal is left **stuck in the scrollback-less alt buffer** until the next tab-switch replay: exactly the bug this PR fixes, recurring intermittently. At ~8 bytes per sequence against ~4KB chunk boundaries the per-occurrence odds are small, but codex re-emits these on every repaint, so over a long session a leak is near-certain. Fixed with a small carry: a trailing partial digit-only CSI intro (≤7 chars) is held back and prepended to the next chunk; complete sequences are never held, and the carry resets in `_resetBuffers`. Two tests added.

**2. Replay didn't strip mouse-tracking enables.** The live strip removes `?1000`–`?1007`, but the `GET /terminal` replay only stripped alt-screen + `3J` — so a codex buffer persisted *before* this PR still re-hijacks the scroll wheel on every tab switch. Added `MOUSE_TRACKING_PATTERN` to the replay strip for parity.

Verified in a clean worktree: `tsc`, `eslint`, `prettier --check` clean; full `test:ci` green (2737 passed, +2 new tests).

PR #118, opened 2026-06-10
feat(terminal): snapshot-replay on tab switches (xterm serialize + live pane capture)

aakhter:
## What

Switching away from a session and back replayed only the server's accumulated byte history. For TUI modes — **codex** especially — that shows just the latest repaint (the idle welcome banner), because the TUI drops earlier conversation from its current frame. This restores the actual on-screen view the user was looking at.

Two complementary mechanisms:

### Client — xterm serialize snapshots
- Loads xterm's `SerializeAddon` and snapshots the rendered state (viewport + scrollback + colors/attrs) per session on switch-**away**, restoring it for an instant first paint on switch-**back**.
- The snapshot is **only the first paint** — the canonical `/terminal` frame is still fetched and reconciled afterward (`restoredSnapshot`/`clearedForBusy` force the replay so a snapshot/busy-clear can't leave stale content even when the fetched buffer byte-matches the cache).
- Snapshots are **LRU-bounded** in memory (≤20) and persisted to `localStorage` (≤256KB each, ≤10 sessions, stale-pruned) so they survive Chrome tab-discard / reload.
- Blank/garbage snapshots are rejected (`_isUsableXtermSnapshot`) before save and restore.

### Server — live pane prepend
- `GET /api/sessions/:id/terminal` now prepends the **live tmux pane buffer** (via the existing `captureActivePaneBuffer`) ahead of the byte history, with a viewport clear between them, so the replay reflects the current frame. No-ops cleanly when no live frame is available (falls back to history).

### Fix: `formatPaneSnapshot` rightmost-column drop
- It painted each captured row to `cols - 1`, dropping the rightmost column. That was caution about last-column autowrap — but every painted row is immediately followed by an absolute cursor-position CSI (the next row's `\x1b[r;1H`, or the final cursor move), which cancels xterm's pending-wrap before any further glyph. So painting the full width is safe; the last column is no longer lost.

## Vendoring
`@xterm/addon-serialize` is added as a pinned dependency and built into `vendor/xterm-addon-serialize.min.js` by `postinstall.js` (dev) and `build.mjs` (prod) — the same esbuild-from-`node_modules` path used for the other xterm addons, so the bundle is reproducible rather than a committed binary.

## Tests
- `test/tmux-manager.test.ts` — `formatPaneSnapshot` re-pinned for full-width painting + a new rightmost-column-preserved case.
- `test/routes/session-routes.test.ts` — `GET /terminal` prepends the live pane (cleared) before history; falls back to history when no live frame.
- `test/codex-snapshot-replay.test.ts` — snapshot save/restore structure: usability gate, first-paint-then-fetch, forced replay after busy-clear, LRU/localStorage bounds.

Verified in a clean worktree against current `master`: `tsc`, `eslint`, `prettier --check`, `check:frontend-syntax`, `check:lockfile`, `check:public-assets` all clean; production `build` regenerates the vendor bundle; full `test:ci` green (2831 passed).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "@fastify/static": "^9.1.3",
         "@fastify/websocket": "^11.2.0",
         "@xterm/addon-fit": "^0.11.0",
+        "@xterm/addon-serialize": "^0.14.0",
         "@xterm/addon-unicode11": "^0.9.0",
         "@xterm/addon-webgl": "^0.19.0",
         "@xterm/xterm": "^6.0.0",
@@ -4523,6 +4524,12 @@
       "version": "0.11.0",
       "resolved": "https://registry.npmjs.org/@xterm/addon-fit/-/addon-fit-0.11.0.tgz",
       "integrity": "sha512-jYcgT6xtVYhnhgxh3QgYDnnNMYTcf8ElbxxFzX0IZo+vabQqSPAjC3c1wJrKB5E19VwQei89QCiZZP86DCPF7g==",
+      "license": "MIT"
+    },
+    "node_modules/@xterm/addon-serialize": {
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@xterm/addon-serialize/-/addon-serialize-0.14.0.tgz",
+      "integrity": "sha512-uteyTU1EkrQa2Ux6P/uFl2fzmXI46jy5uoQMKEOM0fKTyiW7cSn0WrFenHm5vO5uEXX/GpwW/FgILvv3r0WbkA==",
       "license": "MIT"
     },
     "node_modules/@xterm/addon-unicode11": {
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "@fastify/static": "^9.1.3",
     "@fastify/websocket": "^11.2.0",
     "@xterm/addon-fit": "^0.11.0",
+    "@xterm/addon-serialize": "^0.14.0",
     "@xterm/addon-unicode11": "^0.9.0",
     "@xterm/addon-webgl": "^0.19.0",
     "@xterm/xterm": "^6.0.0",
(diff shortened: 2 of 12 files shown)


Ark0N:
Reviewed + pushed fixes to the branch (`9893a7f`). Static checks (tsc, eslint, prettier, frontend-syntax) and the 3 touched test files (99 tests) are green.

**Fixed:**
- **localStorage persistence deadlock (was the main concern):** the old prune only evicted snapshots for sessions that no longer existed, *and* ran only after a successful `setItem`. At the repo's 20-session target (20×256KB ≈ 5MB, right at a typical quota) that meant once the quota filled, `setItem` threw, the catch swallowed it, and the prune was unreachable — persistence silently died forever. Extracted `_persistXtermSnapshot`: evict to a fixed key budget regardless of liveness, plus evict-and-retry inside the catch.
- **FIFO→LRU:** `Map.set` on an existing key doesn't reorder, so the 'oldest' evicted could be the most-active session. Now delete-before-set (save path + localStorage hoist).
- **Shell sessions:** restore is gated on `mode!=='shell'`, but save wasn't — they burned a `serialize()` + cache slot + 256KB localStorage each on every switch-away. Now skipped at save time.
- Tightened the OSC-strip regex in `_isUsableXtermSnapshot` to stop at ST.

**Not changed (intentionally) — follow-up worth considering:** the post-fetch `needsRewrite = restoredSnapshot || …` forces a full clear+rewrite after *every* snapshot restore. The end-state is always canonical, but it (a) reintroduces the documented clear→blank→repaint flash on every idle tab switch, and (b) for codex, replaces the snapshot's clean conversation scrollback with the byte-history replay moments later. Dropping the unconditional `restoredSnapshot ||` would keep the snapshot when `data.terminalBuffer === cachedBuffer` — but per CLAUDE.md that needs real-mobile-device verification (the reverted cross-device buffer-reload is the cautionary precedent), so I left it for a follow-up rather than change redraw behavior blind.

PR #119, opened 2026-06-11
Add server-side attachment pipeline (registry, magic-link, path guard)

aakhter:
## What

Adds a small server-side pipeline that lets a local file be exposed to the browser as a live **external attachment** with a stable id, so requests never carry arbitrary absolute paths.

### Pieces
- **Attachment registry** (`src/attachment-registry.ts`) — in-memory, session-scoped. `registerExternalAttachment` validates an absolute path, resolves symlinks, enforces the path guard, and mints an `att_<uuid>` id. Records are dropped when a session is removed.
- **Path guard** (`src/config/attachment-guard.ts`) — a configurable blocklist (well-known secret locations + the `/root` and `/etc` trees, extendable via the `attachmentBlockedPaths` setting / `CODEMAN_ATTACHMENT_BLOCKED_PATHS`) plus an optional, default-off workspace-confinement mode. The secret-location blocklist is extracted into a shared `src/web/sensitive-path.ts` module, and `/api/download` is refactored to consume it instead of its inline copy (single source of truth).
- **Terminal magic links** — a session scans output for `codeman://attach?path=...` and emits `attachmentRequested`; the web server registers the file and broadcasts an `attachment:detected` SSE event. `codeman attach <path>` prints the magic link (or POSTs directly when a session id is known).
- **Watcher detection** — the image watcher now also detects `png/pdf/docx/pptx` dropped into a session's working directory and emits `attachment:detected`.
- **Routes** — `POST /api/sessions/:id/attachments` (register) and `GET /api/sessions/:id/attachments/:attachmentId/raw` (serve), both re-checking the guard before streaming (defense-in-depth).

## Scope

This is the foundation only. Document previews/thumbnails and the attachment-history drawer build on top of the registry and are intentionally left out of this PR to keep it focused and reviewable.

## Testing

- `tsc --noEmit`, eslint, prettier, frontend-syntax all clean
- Full unit/integration sweep green (2846 passed)
- New/updated tests: attachment magic-link parsing, the path-guard register/serve behaviour (sensitive-file rejection, cross-workspace allow, `/root` + `/etc` defaults, symlink resolution), and image-watcher attachment detection
- Server boot smoke: `/api/status` returns 200

### DIFF
--- a/docs/security-architecture.md
+++ b/docs/security-architecture.md
@@ -320,7 +320,37 @@ injected from API JSON (`innerHTML`), not via `file-raw`, so they are unaffected
 `/api/download` additionally refuses a blocklist of sensitive paths
 (`/etc/shadow`, `~/.ssh/`, `.env`, `*credentials*`, `.aws/credentials`, …). This
 is **defense‑in‑depth, not the primary boundary** — the realpath containment is
-the control.
+the control. The blocklist patterns are shared (`src/web/sensitive-path.ts`) with
+the attachment guard below.
+
+### External attachments (registry) & the magic‑link trust boundary
+
+Live external attachments (`src/attachment-registry.ts`) mint an `att_<uuid>` id
+for a host file so browser requests carry the id, never an absolute path. Serving
+is by id (`GET /api/sessions/:id/attachments/:attachmentId/raw`, 50 MB cap,
+`nosniff`) and re‑resolves the symlink + re‑checks the **attachment guard**
+(`src/config/attachment-guard.ts`: the shared sensitive‑path blocklist **plus**
+the `/root` and `/etc` trees, extendable via `attachmentBlockedPaths` /
+`CODEMAN_ATTACHMENT_BLOCKED_PATHS`) on every request. Unlike the workspace file
+routes, attachments are intentionally **cross‑workspace** — so the effective gate
+is the blocklist + a 6‑extension allowlist (`png/pdf/docx/pptx/md/txt`), not
+realpath containment.
+
+Two registration paths, with **different trust**:
+
+- **Explicit `POST /api/sessions/:id/attachments`** (and `codeman attach`, which
+  POSTs directly inside a managed session) — a deliberate, Origin‑guarded HTTP
+  request. Allowed cross‑workspace (subject to the guard). This is the supported
+  path for codeman‑publish and the `~/.codeman` review‑card loop.
+- **Terminal `codeman://attach?path=…` magic links** — scanned passively from
+  session output. Terminal output is **attacker‑influenceable** (a prompt‑injected
+  session can print an arbitrary path), and registration here is server‑side with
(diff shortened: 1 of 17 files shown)


Ark0N:
Reviewed (security focus) + pushed fixes, rebased onto current master (`f7ce8e4`). tsc / eslint / frontend-syntax / format:check clean; the 3 touched test suites pass (56 tests, +2 new for forced confinement).

### Blockers fixed

**1. Magic-link injection → arbitrary host-file read (security).** The terminal `codeman://attach?path=…` scanner registered any matching path **server-side, with no user confirmation and no Origin gate**, then broadcast the `rawUrl` over SSE to every client. Terminal output is attacker-influenceable — the repo's own threat model names a prompt-injected session printing arbitrary output as a realistic actor — so on the default no-auth loopback + `tailscale serve` deployment this was an arbitrary host-file (`png/pdf/docx/pptx/md/txt`) read primitive reachable by any tailnet SSE client; the only gate was the extension allowlist.

Fix: the magic-link scan path is now **force-confined to the session workspace** (`forceWorkspaceConfinement` in `registerExternalAttachment`, wired in `WebServer.registerAttachment`), regardless of the global `attachmentConfineToWorkspace` setting. **Deliberate cross-workspace attach is unaffected** — it goes through the explicit, Origin-guarded `POST /attachments` and `codeman attach`, which POSTs directly inside a managed session (`CODEMAN_SESSION_ID` is exported, so the CLI never relies on the scannable magic link for cross-workspace files). So codeman-publish + the `~/.codeman` review-card loop keep working; only passive prompt-injectable links are confined. Documented in `docs/security-architecture.md §5`.

  > Note: this confines the contributor's intended cross-workspace *magic-link* behavior. If you specifically want raw terminal output (not the CLI/POST) to attach out-of-workspace files, the alternative is a UI approval gate (broadcast a 'pending' card the user clicks) — happy to do that as a follow-up, but force-confine is the safe default.

**2. PNG popup regression.** `.png` was rerouted from `image:detected` to `attachment:detected`, which has **no frontend consumer** (the attachment UI is out of scope for this PR) — silently breaking the dropped/pasted-screenshot popup. PNG stays on `image:detected`; only `pdf/docx/pptx` (no prior popup) emit `attachment:detected`.

### Minors fixed
- **TOCTOU:** the raw route now streams the freshly-resolved path, not the stored one, so a post-registration symlink swap can't redirect the stream to a non-blocklisted target.
- 50 MB cap on the attachment raw route (matched `file-raw`/`download`).
- Per-session registry cap (200) to bound the POST path.
- CLI reads creds via `dataPath('.env')`, honoring `CODEMAN_INSTANCE`.

(Skipped the suggested async-realpath change — the test harness mocks only the sync `node:fs` realpath, the path is already bounded 200/session, and the rest of the codebase uses sync realpath here, so it'd be churn for a marginal perf win.)

PR #120, opened 2026-06-12
Document attachment previews + thumbnails (attachment cards)

aakhter:
Builds on the merged attachment registry (#119) to add the **attachment-card consumer** that PR noted was out of scope: detected/registered attachments now surface as dismissible cards with a first-page thumbnail and an inline preview.

### Backend
- **`document-thumbnailer`** — first-page PNG thumbnails: PNG passthrough, PDF via `pdftoppm`, Office via the preview cache.
- **`document-preview-cache`** — disk-cached DOCX/PPTX → PDF conversion (LibreOffice, with a PowerShell-COM fallback on Windows), in-flight dedup, multi-converter fallback.
- **`file-routes`** — `serveConvertedPreview` / `serveThumbnail` + four routes: `GET /api/sessions/:id/attachments/:attachmentId/preview` and `.../thumbnail`, plus the workspace-path `file-preview` / `file-thumbnail`. The by-id routes reuse the registry's TOCTOU-safe `resolveServableAttachmentPath`, so previews stream the freshly-resolved path.
- **`server`** — enriches detected attachment events with a thumbnail route so cards show a thumbnail.
- **`image-watcher`** — `.png` now routes to `attachment:detected`. The previous PR kept PNG on `image:detected` because there was no card consumer; this PR adds that consumer, so PNG screenshots become cards (the popup path stays for jpg/gif/webp/etc.).

### Frontend
- **`panels-ui`** — attachment cards (`addAttachmentCard`, a lazily-created floating stack, a "Clear all" control that appears at ≥2 cards, per-session cleanup on close) and a backward-compatible 3-arg `openFilePreview` that renders registered attachments inline (image / PDF) or via the server-converted PDF (docx/pptx).
- **`app.js`** — wires `attachment:detected` → `_onAttachmentDetected` and the card state.
- **`styles`** — attachment-card + stack styling.

### Verification
`tsc`, `eslint`, `prettier --check`, `check:frontend-syntax`, `check:public-assets` all clean. New `document-thumbnailer` + `document-preview-cache` unit tests pass; full `test:ci` green (2861 passed). Card render, preview overlay, and dismiss verified in-browser.

PR #121, opened 2026-06-12
Attachment history drawer

aakhter:
**Stacked on #120** (document attachment previews). Until #120 merges, this PR's diff also shows the COD-38 commit (`49c92e4`); the history commit is `3731aef`. Once #120 lands, rebasing leaves only the history change.

Accumulates a per-session attachment history and exposes it through a slide-in drawer with an unread badge, so attachments stay reachable after their cards are dismissed.

### Backend
- **`session-attachment-history`** — history state: dedupe by source/relative path, newest-first, 100-item cap, and `externalPath` sanitization (the absolute host path is server-private and never leaves `toState()`).
- **`session.ts`** — `_attachmentHistory` + sanitized getter / `upsertAttachmentHistory` / `restoreAttachmentHistory` / `getAttachmentHistoryForPersist`; restored from saved state in the constructor.
- **`file-routes`** — `GET /api/sessions/:id/attachments` (list — resolves each entry to live metadata + routes; external entries are re-registered so the guard runs again) and `GET /api/sessions/:id/attachments/:attachmentId` (metadata poll, guarded by the registry's TOCTOU-safe `resolveServableAttachmentPath`).
- **`server.ts`** — detected/registered attachments upsert into history and persist; the private (`externalPath`-bearing) history rides on disk under `__attachmentHistory`, separate from the sanitized public copy in session state, and is restored on mux-session recovery.
- **`types/session.ts`** — `SessionAttachmentHistoryItem` + `SessionState.attachmentHistory`.

### Frontend
- **`panels-ui`** — the drawer (lazily built, appended to `<body>`), the unread badge, list rendering with per-item Preview / Download / Open / "Card" (reshow) actions, and a debounced live refresh of the open drawer on new detections.
- **`app.js`** — history state + per-session badge/cleanup wiring (clear on kill-all, refresh on tab switch, drop on session removal).
- **`index.html` / `styles.css` / `mobile.css`** — header button + badge and the drawer (full-screen sheet on phones).

### Verification
`tsc`, `eslint`, `prettier --check`, `check:frontend-syntax`, `check:public-assets` clean. New history-module unit tests pass; full `test:ci` green (2866 passed). Badge bump, drawer open/render/reshow/close verified in-browser.

### DIFF
--- a/src/document-preview-cache.ts
+++ b/src/document-preview-cache.ts
@@ -1,0 +1,256 @@
+/**
+ * @fileoverview Shared disk cache for expensive Office document previews.
+ */
+
+import { createHash } from 'node:crypto';
+import { execFile } from 'node:child_process';
+import fs from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { basename, dirname, extname, join } from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { promisify } from 'node:util';
+
+const execFileAsync = promisify(execFile);
+
+const OFFICE_CONVERSION_TIMEOUT_MS = 5 * 60_000;
+const DOCUMENT_PREVIEW_CACHE_DIR = join(tmpdir(), 'codeman-document-preview-cache');
+function buildWordExportPdfScript(sourcePath: string, outputPath: string): string {
+  return `
+$ErrorActionPreference = "Stop"
+$source = ${toPowerShellSingleQuotedString(sourcePath)}
+$output = ${toPowerShellSingleQuotedString(outputPath)}
+$word = $null
+$doc = $null
+try {
+  $word = New-Object -ComObject Word.Application
+  $word.Visible = $false
+  $word.DisplayAlerts = 0
+  $doc = $word.Documents.Open($source)
+  $doc.ExportAsFixedFormat($output, 17)
+} finally {
+  if ($null -ne $doc) {
+    $doc.Close($false) | Out-Null
+    [System.Runtime.InteropServices.Marshal]::ReleaseComObject($doc) | Out-Null
+  }
+  if ($null -ne $word) {
+    $word.Quit() | Out-Null
+    [System.Runtime.InteropServices.Marshal]::ReleaseComObject($word) | Out-Null
+  }
+  [System.GC]::Collect()
+  [System.GC]::WaitForPendingFinalizers()
+}
+`.trim();
+}
+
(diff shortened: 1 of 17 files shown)


Ark0N:
**Merged into `master` 🎉** (landed via rebase as commit `577b6d7` — your authorship preserved — plus a review-fixes commit `1a363a3`, merged in `e742d00`). Because it was rebased onto the post-#120 master, GitHub shows this as *closed* rather than the purple *merged* badge, but it's in master.

Thanks for this, @aakhter — solid feature. I ran a deep multi-agent review (23 findings, **all minor/nit, zero blockers**) and verified the security posture: `externalPath` never leaves the server — the sanitized `toState()`, the list route's `Omit<…,'externalPath'>` shape, and re-registration-through-the-guard all hold. Persist/restore round-trips correctly.

Heads-up: the committed `styles.css` ended mid-rule (`​.attachment-history-actions button:disabled {` with no body/closing brace — prettier doesn't flag a trailing unclosed block, so CI was green and the browser tolerated it). I completed it during the rebase.

Review fixes I applied (separate commit, all tested):
- **Recovery guard:** `restoreAttachmentHistory` now skips malformed/legacy saved items instead of throwing in the `Session` constructor (a corrupt `__attachmentHistory` entry could otherwise abort the whole mux-recovery loop).
- **Resilient list route:** one failing entry degrades to `{missing:true}` instead of failing the whole drawer.
- **Badge anchoring:** gave the header button a positioning context so the badge pins to the icon, not the header bar.
- **Debounce:** cancelled on drawer-close + guarded against a stale session.
- **Re-show:** uses the item's own timestamp so the cardId is stable (focuses the existing card instead of duplicating).
- **Escape** now closes the drawer; badge shows `99+`; dropped a duplicate `@keyframes`; fixed 3 undefined CSS custom properties in the empty-state.
- Added constructor-restore round-trip + malformed-item tests.

Verified: `tsc`, eslint, prettier, frontend-syntax, full `test:ci` (2906 pass), and a real-browser check (badge positioning, drawer open, Escape-close).

Deferred for your call (none correctness/security): the full 100-item history is broadcast in every session-state SSE event (payload bloat — worth stripping from the light serializer); the badge is a total-count labeled 'unread'; the header button is always-on vs the opt-in convention of its neighbors; and `app.inject` route tests for the two new endpoints. Happy to do these as a follow-up if you'd like.

PR #124, opened 2026-06-13
feat(settings): add Opus 4.6 model options

TeigenZhang:
## Summary
- Add `claude-opus-4-6[1m]` (1M context) and `claude-opus-4-6` to the Claude Model picker dropdown

## Test plan
- [ ] Open App Settings → Claude Model dropdown shows the two new Opus 4.6 entries
- [ ] Select Opus 4.6 (1M context) → creates session with correct model

### DIFF
--- a/src/web/public/index.html
+++ b/src/web/public/index.html
@@ -1166,6 +1166,8 @@
                 <option value="claude-fable-5">Fable 5</option>
                 <option value="opus[1m]">Opus (1M context)</option>
                 <option value="opus">Opus</option>
+                <option value="claude-opus-4-6[1m]">Opus 4.6 (1M context)</option>
+                <option value="claude-opus-4-6">Opus 4.6</option>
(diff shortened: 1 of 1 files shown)


Ark0N:
Thanks for the contribution, @TeigenZhang! 🙌

Verified end-to-end before merging:
- The dropdown values flow `claudeModel` → `modelOverride` → `updateCaseModel()`, which writes the string verbatim into the case's `.claude/settings.local.json` as `model`. The `[1m]` suffix is interpreted by Claude Code itself (same proven pattern as the existing `opus[1m]` / `claude-fable-5[1m]` entries), so the new options are pure pass-throughs — no schema or allowlist change needed.
- `claudeModel`/`modelOverride` are free-form `z.string().max(50)`, and both new values fit comfortably.
- CI is green (typecheck/lint + full test suite), and `format:check` covers `index.html`.

Clean, well-scoped change. Merging. 🚀

PR #138, opened 2026-07-01
Raise terminal history/scrollback/buffer defaults (50k→100k, 2MB→32MB)

aakhter:
## What

Raises the centralized terminal-history defaults introduced in #137:

| value | before (#137) | this PR |
|---|---|---|
| tmux `history-limit` | 50,000 | 100,000 |
| terminal scrollback lines | 50,000 | 100,000 |
| PTY buffer max | 2 MB | 32 MB |
| PTY buffer trim | 1.5 MB | 24 MB |

All values remain env- and settings-overridable (`CODEMAN_MAX_TERMINAL_BUFFER` / `CODEMAN_TRIM_TERMINAL_TO`) and bounds-clamped via `resolveTerminalHistoryConfig()`.

## Why split from #137

#137 introduced the config mechanism with **no behavior change** (defaults kept at the prior hardcoded values), so the plumbing could be reviewed on its own. This PR is the follow-up policy change: larger defaults so a full default scrollback is retained and replayable. Keeping them separate lets the larger memory budget be weighed independently.

## Trade-off

Worst-case terminal-buffer budget for 20 concurrent sessions rises **40 MB → 640 MB** (comment updated to match). Operators who want the smaller footprint can set `CODEMAN_MAX_TERMINAL_BUFFER` / `CODEMAN_TRIM_TERMINAL_TO`.

## Tests

`test/terminal-history.test.ts` updated to assert the raised defaults; `tsc --noEmit`, the terminal-history + schema tests, and `npm run build` all pass.

### DIFF
--- a/src/config/buffer-limits.ts
+++ b/src/config/buffer-limits.ts
@@ -6,10 +6,9 @@
  * it easy to tune memory usage.
  *
  * Memory Budget Rationale (for 20 concurrent sessions):
- * - Terminal buffer: 2MB max × 20 = 40MB worst case
+ * - Terminal buffer: 32MB max × 20 = 640MB worst case
  * - Text output: 1MB max × 20 = 20MB worst case
  * - Messages: ~1KB each × 1000 × 20 = 20MB worst case
- * - Total buffer overhead: ~80MB (acceptable for a long-running server)
  *
  * @module config/buffer-limits
  */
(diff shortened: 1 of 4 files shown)


Ark0N:
Merged — thank you @aakhter! The raised tmux history (100k) and PTY buffer caps (32MB/24MB) land in the next release. During review we appended one commit: the env-path trim default is now clamped below the max (an operator setting only `CODEMAN_MAX_TERMINAL_BUFFER` could otherwise disable trimming entirely), and the scrollback-lines constant was scoped back to what's actually wired so the docs don't overclaim. Appreciate the steady stream of solid infra work you've been sending in! 🙏

PR #139, opened 2026-07-01
Unified session list: backend service + endpoint

aakhter:
First increment of a read-only "complete + searchable session list" — the backend half. Frontend (rich list UI, persistent modal, live refresh) and polish (pinning, cross-device ordering, name/prompt retention) follow in stacked PRs.

### What this adds
- **New `src/services/unified-session-service.ts`** — pure, IO-free, unit-testable:
  - `mergeUnifiedSessions()` combines five sources — live sessions + persisted (`state.json`) + lifecycle log + `~/.claude` transcript history + mux stats — into one list de-duped by `sessionId`, with precedence `history < lifecycle < persisted < live`, a meaningfulness floor that drops bare lifecycle/mux-only noise, and a stable newest-first sort.
  - `filterAndPaginate()` — case-insensitive `q` over name/firstPrompt/workingDir/sessionId; `total` computed before paging; `limit` clamped to `[1, 500]`.
- **New `GET /api/sessions/unified`** in `session-routes.ts` — gathers the five sources from ctx (each `try`/caught so one bad source can't fail the request), feeds the pure service, returns `{ sessions, total }` in the standard `ApiResponse` envelope. Short-circuits to empty in test mode.

### Tests
- `test/services/unified-session-service.test.ts` — 12, pure (merge precedence, de-dup, meaningfulness floor, sort, filter, pagination, limit-clamp).
- `test/routes/unified-sessions-routes.test.ts` — 4, via `app.inject` (envelope, pagination, q-search, limit-clamp).

### Verification
`tsc --noEmit` clean · new tests 16/16 · existing `session-routes` suite 53/53 (no regression) · prettier + eslint clean · build clean.

Read-only and additive — no existing route or behavior changes.

### DIFF
--- a/src/services/unified-session-service.ts
+++ b/src/services/unified-session-service.ts
@@ -1,0 +1,235 @@
+/**
+ * @fileoverview Pure merge/filter logic for the unified session list (COD-121).
+ *
+ * Combines four read-only views of a session — live (in-memory `Session`),
+ * persisted (`state.json`), transcript history (`~/.claude/projects`), and the
+ * lifecycle audit log — plus mux process stats, into one de-duplicated list
+ * keyed by sessionId. Higher-precedence sources overwrite scalar fields when
+ * present (history < lifecycle < persisted < live), while the `sources` array
+ * always accumulates every contributing view. A "meaningfulness floor" drops
+ * noise (bare lifecycle/mux-only rows with no name and no first prompt).
+ *
+ * PURE: no fs/IO and no node imports. All IO happens in the route that feeds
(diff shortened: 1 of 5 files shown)


Ark0N:
Merged — thank you @aakhter! The unified session service is a great foundation (and PR #146's palette plugs straight into it). Review appended one commit: resumed/respawned sessions now dedupe through a claudeSessionId→session-id alias map, and lifecycle name/mode resolution is newest-wins (query() returns newest-first, so the old loop let the oldest entry win). Both covered with tests. 🙏

PR #140, opened 2026-07-02
feat(terminal): add WebGL renderer toggle in settings

kvncrw:
Adds a 'WebGL Renderer' toggle to Settings > Appearance (desktop). WebGL stays on by default; users can turn it off to force the DOM renderer when they hit GPU glitches, without needing the ?nowebgl URL param. Explicit opt-in (or ?webgl=force) clears a stale auto-fallback marker. Mobile skip and the long-task auto-fallback safety net are unchanged.

The device/param/sticky/pref interaction is factored into a pure, unit-tested shouldSkipWebGL() helper in constants.js.

### DIFF
--- a/.changeset/webgl-renderer-toggle.md
+++ b/.changeset/webgl-renderer-toggle.md
@@ -1,0 +1,5 @@
+---
+"aicodeman": patch
+---
+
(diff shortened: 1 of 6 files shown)


Ark0N:
Merged — thank you @kvncrw, and welcome as a first-time contributor! 🎉 The WebGL toggle is genuinely useful for GPU-glitchy setups. Two notes: (1) your org fork (crawlsys) rejects maintainer pushes, so we merged locally with the review fixes included — the PR shows as merged via the master merge commit; (2) the review commit makes the setting per-device (renderer choice is GPU-specific, and the strict settings schema would otherwise have 400'd every save) and preserves the sticky auto-fallback marker unless the user explicitly re-enables. Hope to see more from you!

PR #143, opened 2026-07-07
fix(mobile): CJK input loss — IME state machine, focus routing, and Android InputConnection recovery

TeigenZhang:
## Summary

Fixes Chinese/Japanese/Korean (CJK) input silently disappearing on mobile — characters composed in the IME never reach the PTY, or land in the wrong place after switching apps. Three distinct root causes, one per layer:

1. **IME state machine** — `compositionstart`/`compositionupdate`/`compositionend` were not tracked as a proper state machine, so a composition that was interrupted (app switch, focus steal) left `cjkActive` stuck or cleared at the wrong time, dropping the committed text.
2. **Focus routing** — session-select and SSE-reconnect restore paths call `terminal.focus()`, which stole focus back to xterm's hidden textarea while the CJK field was visible. The IME then composed into a black hole (xterm's `onData` is gated by `cjkActive`). Now every `terminal.focus()` routes through one chokepoint that keeps the CJK field focused when it's visible.
3. **Android InputConnection recovery** — some Android keyboards tear down and rebuild the `InputConnection` mid-composition; the textarea silently stopped emitting input events. Detect the dead connection and re-establish it.

## Changes

- `input-cjk.js` — composition state machine + Android InputConnection recovery
- `terminal-ui.js` — focus router chokepoint so restore paths don't steal focus from the CJK field
- `app.js` / `server.ts` — crash-diag tracing hooks used to pin down the Android case
- `test/input-cjk.test.ts` — 9 new cases (241 lines) covering the state machine and focus routing

## Test plan

- [x] `npm test -- test/input-cjk.test.ts` — 9/9 pass
- [x] `npm run check:frontend-syntax`
- [x] Verified on a real Android device (the InputConnection path does not reproduce in Chromium/Playwright)


Ark0N:
Merged — thank you @TeigenZhang! The IME state machine + Android InputConnection recovery is careful, well-documented work — CJK input on mobile is dramatically better for it. Review appended one commit: diagnostic traces are now content-free (the crash-diag pipeline persists locally and server-side, so raw keystrokes/textarea values couldn't ride along), the onData self-heal only refocuses on genuine typed input (xterm's self-generated DA/DSR query replies were stealing focus from other inputs), and the tap-to-unwedge recovery is gated to Android where the wedge exists. Your continued mobile/CJK work is hugely appreciated! 🙏

PR #144, opened 2026-07-07
feat(web): restore tap/click/wheel mouse interaction when server strips mouse DECSETs

TeigenZhang:
## Summary

Restores mouse interaction with the terminal (tap-to-position cursor, desktop click-to-position, and scroll-wheel forwarding) for sessions whose output the server strips of mouse-tracking DECSETs.

**Root cause.** To keep the wheel scrolling scrollback, the server strips `?1000/1002/1003/1006` DECSET sequences from claude/codex/gemini output (`isAltScreenStripMode`, `session.ts`). Side effect: the browser xterm's `mouseTrackingMode` is permanently `'none'` for those sessions, even though the TUI on the PTY side still has tracking ON and understands SGR reports. Any UI that gates on xterm's live tracking mode — mobile touchend tap, desktop click, wheel forwarding — silently does nothing.

**Fix.** When a session is one the server strips, hand-encode the SGR report from the pointer position and send it straight to the PTY, bypassing xterm's dormant encoder:

- **Mobile tap** — touchend tap encodes an SGR press+release (was already partly handled; consolidated here).
- **Desktop click** — a plain left-click encodes SGR press+release. Every click that already means something else is skipped: synthetic/compat clicks after a touch tap, modified clicks (Shift keeps xterm's selection override), double/triple clicks (word/line select), drag-selections, off-grid clicks, and sessions where xterm's own encoder is live.
- **Wheel** — Claude Code 2.1.187+ scrolls its own transcript on SGR wheel reports and no longer captures wheel as select-menu navigation (verified against 2.1.202: the `/model` menu highlight ignores wheel reports). So the wheel is forwarded to the TUI for strip-mode sessions at the buffer bottom, coalesced into one PTY write per ~40ms to avoid a `send-keys` process storm. `Shift+wheel` and any scrolled-up viewport stay on xterm's local scrollback.

**Scrollback coordinate guard.** SGR coordinates are viewport-relative, so a report computed while scrolled up would hit-test the wrong row. Synthetic taps/clicks/wheel are gated on the viewport being at the buffer bottom.

Behavior is unchanged for shell/opencode sessions and when a session genuinely has mouse tracking off.

## Changes

- `terminal-ui.js` — `_handleDesktopTerminalClick`, `_sendSyntheticSgrTap`/`Wheel`, `_shouldForwardWheelToApp`, `_terminalViewportAtBottom`, `_clientPointToCell`; wheel handler forwards to the app for strip-mode sessions
- `test/terminal-touch-tap.test.ts` — 12 cases (SGR encoding, grid clamping, click-skip conditions, wheel gating/coalescing, scrollback guard, shell-mode exclusion)

## Test plan

- [x] `npm test -- test/terminal-touch-tap.test.ts` — 12/12 pass
- [x] `npm run check:frontend-syntax`
- [x] Verified E2E via Playwright against a live instance: desktop click emits SGR press+release; wheel up/down forward SGR 64/65; `Shift+wheel` stays on local scrollback
- [x] Verified the TUI side by injecting SGR bytes into a real claude session in an isolated tmux: click expands a collapsed tool-output block; wheel scrolls the transcript; the `/model` select menu ignores wheel reports

### DIFF
--- a/src/web/public/index.html
+++ b/src/web/public/index.html
@@ -560,6 +560,7 @@
             <div><kbd>Alt/Option</kbd>+<kbd>[</kbd> / <kbd>Alt/Option</kbd>+<kbd>]</kbd></div><div>Previous / Next Session</div>
             <div><kbd>Alt/Option</kbd>+<kbd>1-9</kbd></div><div>Switch to Tab N</div>
             <div><kbd>Ctrl</kbd>+<kbd>L</kbd></div><div>Clear Terminal</div>
+            <div><kbd>Shift</kbd>+<kbd>Wheel</kbd></div><div>Scroll local history (when mouse passthrough is active)</div>
             <div><kbd>Ctrl</kbd>+<kbd>+</kbd></div><div>Increase Font</div>
             <div><kbd>Ctrl</kbd>+<kbd>-</kbd></div><div>Decrease Font</div>
             <div><kbd>Ctrl</kbd>+<kbd>?</kbd></div><div>Show Help</div>
--- a/src/web/public/terminal-ui.js
+++ b/src/web/public/terminal-ui.js
@@ -323,13 +323,25 @@ Object.assign(CodemanApp.prototype, {
     // Register link provider for clickable file paths in Bash tool output
     this.registerFilePathLinkProvider();
 
-    // Always use mouse wheel for terminal scrollback, never forward to application.
-    // Prevents Claude's Ink UI (plan mode selector) from capturing scroll as option navigation.
+    // Mouse wheel: forward to the TUI for strip-mode sessions, local scrollback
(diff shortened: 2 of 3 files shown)


Ark0N:
Merged — thank you @TeigenZhang! Restoring tap/click/wheel over the server-side DECSET strip closes a real interaction gap. Review appended one commit: wheel-forwarding is now version-gated (claude ≥ 2.1.187 — older Ink select menus capture wheel as option navigation, the exact bug the strip guarded against; unknown versions fall back to local scroll), gemini is excluded until its TUI behavior is verified, clicking a file-path/URL link no longer double-fires a synthetic click into the TUI, and Shift+wheel (local history) is documented in the shortcuts modal. 🙏

PR #147, opened 2026-07-09
fix(session): scrub TMUX vars + PTY-exit circuit breaker (COD-115/COD-118)

aakhter:
## Summary

Two related fixes for tmux-backed session stability:

**COD-115 — Delete TMUX/TMUX_PANE (don't set-to-undefined)**

When the Codeman server is launched from *inside* a tmux session, `process.env.TMUX` is set. The mux-attach PTY inherits that env, and tmux detects the nesting via `TMUX`/`TMUX_PANE` → exits 1 → Codeman's respawn kicks in → crash-loop.

Root cause: `buildMuxAttachEnv()` was spreading `{...process.env, TMUX: undefined}`. node-pty serializes that as the literal string `"TMUX=undefined"` (the key is *present* with a non-empty value), so the guard still fires. Fix: `delete env.TMUX` and `delete env.TMUX_PANE` — key absent from the environment entirely. Same fix applied at web bootstrap in `src/index.ts` for belt-and-suspenders coverage.

**COD-118 — PTY-exit circuit breaker**

Introduces `InteractivePtyExitBreaker` (pure class, injectable time for tests): a sliding-window breaker that trips after 5 non-zero PTY exits within 10 seconds, sets `_respawnBlocked`, and emits `respawnBreakerTripped`. SSE event + push notification tell the user. Explicit user restart (`POST /api/sessions/:id/start`) resets the breaker before relaunching.

Prevents a misconfigured case (bad shell, bad Claude install, permission denied) from hammering tmux/PTY in an infinite respawn loop and burning system resources.

## Files changed

- `src/index.ts` — delete TMUX/TMUX_PANE at web bootstrap
- `src/session-cli-builder.ts` — `buildMuxAttachEnv()` uses `delete` for TMUX, TMUX_PANE, CLAUDECODE, COLORTERM
- `test/session-cli-builder.test.ts` — new; COD-115 absence assertions (`'TMUX' in env === false`)
- `src/session-pty-exit-breaker.ts` — new; `InteractivePtyExitBreaker` (sliding window, threshold 5/10s)
- `src/session.ts` — `_ptyExitBreaker` + `_respawnBlocked`, `resetRespawnBreaker()`, PTY exit instrumentation
- `src/web/session-listener-wiring.ts` — `respawnBreakerTripped` handler → SSE + push + run-summary
- `src/web/sse-events.ts` — `SessionRespawnBreakerTripped` event
- `src/web/public/constants.js` — `SESSION_RESPAWN_BREAKER_TRIPPED` constant
- `src/web/public/app.js` — diagnostic toast on `respawnBreakerTripped`
- `src/web/routes/session-routes.ts` — `session.resetRespawnBreaker()` before `startInteractive()`
- `test/respawn-pty-breaker.test.ts` — new; 13 tests (pure breaker + session integration)

## Test plan

- `npm test -- test/respawn-pty-breaker.test.ts test/session-cli-builder.test.ts` → 15/15 pass
- `tsc --noEmit` → clean
- `npm run build` → clean

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,18 @@ import { program } from './cli.js';
 // In web mode, we should NOT exit on transient errors — log and continue
 const isWebMode = process.argv.includes('web');
 
+// COD-115: Codeman IS a tmux controller; it must never present as a tmux *client*.
+// If the web server is launched from inside a tmux pane it inherits TMUX/TMUX_PANE,
+// and tmux's nesting guard then kills every new attach-bridge PTY (exit 1 → respawn
+// loop, crash-looping any new tmux-backed session). Scrub at the root so every
+// downstream `{...process.env}` spread (attach, send-keys, create) is clean regardless
+// of launch context. `delete` (not `= undefined`, which node-pty serializes as the
+// literal string "undefined" and fails to clear).
+if (isWebMode) {
+  delete process.env.TMUX;
+  delete process.env.TMUX_PANE;
+}
+
 import { MAX_CONSECUTIVE_ERRORS, ERROR_RESET_MS } from './config/server-timing.js';
 
 // Track consecutive unhandled errors in web mode — restart after too many
--- a/src/session-cli-builder.ts
+++ b/src/session-cli-builder.ts
@@ -124,17 +124,32 @@ export function buildClaudeEnv(sessionId: string): Record<string, string | undef
  * Lighter than buildClaudeEnv — no PATH augmentation or Codeman vars needed
  * since the mux session already has those set.
  *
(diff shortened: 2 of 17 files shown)


Ark0N:
Merged — thank you @aakhter! TMUX-var scrubbing + the PTY-exit breaker close a real crash-loop gap (COD-115/118). Review adjusted the reset semantics: the breaker now clears only on an explicit `{clearBreaker:true}` restart — the auto-reattach path (every tab click / page load) was resetting it, so it could never trip on the exact crash loop it targets; trip events also survive listener detach and the push notification is wired into PUSH_EVENT_MAP. 🙏

PR #148, opened 2026-07-10
feat(terminal): replay full tmux scrollback on terminal reload + CRLF normalization

aakhter:
## Summary

A full page reload (`GET /api/sessions/:id/terminal` without `?tail=`) previously only replayed Codeman's in-memory byte buffer — scrollback that had been pushed off the buffer was lost. This PR adds two complementary fixes:

### COD-47: Full tmux scrollback replay

- `GET /api/sessions/:id/terminal` (no `?tail=`) now calls `capture-pane -p -e -S -`, capturing the **entire** tmux scrollback, not just the visible frame.
- Tab switches (`?tail=N`) keep using the fast visible-frame path — no regression.
- Shell-mode sessions apply a VPA-strip gate before returning the buffer (skips Ink redraw bloat).
- `MuxInterface.capturePaneBuffer` gains an optional `opts.fullHistory` flag to request the deep capture.

### COD-138: CRLF normalization of scrollback replay

Without this, full-history capture causes a "staircase" rendering bug: `capture-pane -S -` returns scrollback with bare `\n` line endings; xterm.js running with `convertEol: false` (correct for the live PTY stream, which carries real `\r\n`) treats each bare `\n` as "move down without carriage return," shifting each replayed line one column to the right.

Fix: new pure helper `normalizeScrollbackEol()` converts `\r?\n` → `\r\n` (idempotent on CRLF; preserves lone `\r` overwrites; adds/removes no rows) applied at the raw-return seam. The absolute-cursor snapshot path (visible-frame / tab-switch) is unchanged.

## Test plan

- `test/tmux-capture-full-history.test.ts` (new, 5 tests) — full-history flag + VPA-strip gate
- `test/tmux-scrollback-eol.test.ts` (new, 6 tests) — CRLF normalization invariants (no bare LF, CRLF idempotency, lone-CR preservation)
- `test/routes/session-routes.test.ts` (13 new terminal-route tests added, 63 total pass)
- `tsc --noEmit`: 0 errors
- `npm run build`: clean
- Prettier: clean

### DIFF
--- a/src/mux-interface.ts
+++ b/src/mux-interface.ts
@@ -98,6 +98,20 @@ export interface RespawnPaneOptions {
   historyLimit?: number;
 }
 
+/** Options for pane buffer capture (COD-47 full-history mode). */
+export interface PaneCaptureOptions {
+  /** Capture the entire tmux scrollback instead of just the visible frame. */
+  fullHistory?: boolean;
+  /** Bound the full-history capture to this many scrollback lines (`-S -<N>`). */
+  historyLimitLines?: number;
+  /**
+   * Byte cap the consumer will keep from the capture. Sizes the child-process
+   * stdout buffer (with slack) so multi-MB scrollback dumps aren't killed by
+   * the 1MB execSync default (ENOBUFS).
+   */
+  maxCaptureBytes?: number;
+}
+
 /**
  * Terminal multiplexer interface.
  *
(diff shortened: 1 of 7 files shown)


Ark0N:
Merged — thank you @aakhter! Full-scrollback replay makes reloads dramatically more useful. Review wired the trigger end-to-end (an explicit `?full=1` requested by the initial page load — the no-tail heuristic never fired from the real frontend paths), sized execSync's maxBuffer from config (the 1MB default silently killed exactly the big captures the feature exists for), bounded the capture, and returned it alone (the byte-buffer prepend replayed the whole history twice). 🙏

PR #149, opened 2026-07-11
fix(ws): WebSocket durable-delivery resilience — reconnect, per-client scoping, ACK re-drive

aakhter:
## Summary

Six bug-fix commits hardening the WebSocket layer against reconnect races, spurious disconnects, stale ACKs, and redundant DOM work. Each is self-contained and builds cleanly on the previous.

### COD-133 — Fix header WS status indicator + typing lag

Restores the connection-state chip (broken since v1.1.15 merge) and removes a `keydown` handler that was blocking the UI thread on every keystroke.

### COD-134 — Fix WS flap loop + reconnect resilience

- Fixes `undefined onopen call` crash on rapid connect/disconnect cycles that caused an infinite flap loop.
- Adds `_wsLastRecvAt` / `_wsState` tracking so reconnect back-off is state-machine driven rather than timer-soup.
- Exports `CodemanWsReconnect` / `planWsReconnect` from `constants.js` for testability.

### COD-135 — Re-drive lost input ACK on a live WebSocket

When an input frame is sent but the ACK never arrives (connection drop between send and response), the client now re-drives the pending frame on the next live connection rather than silently dropping it. Closes the durable-delivery gap without requiring a page reload.

### COD-137 — Scope WS per-session limit by `clientId`

Upstream's per-session connection cap (max 1 WS per session) was keyed on `sessionId` alone, causing spurious `4008` closes when a client reconnected before the server had cleaned up the old socket. New `WsConnectionRegistry` (`src/web/ws-connection-registry.ts`) scopes the limit by `(sessionId, clientId)` pair so reconnects are never rejected.

### COD-136 — Skip redundant connection-indicator DOM writes on hot input path

`_updateConnectionIndicator()` was called on every received message even when the indicator state hadn't changed. Added a dirty-check; saves several `classList` mutations per second on busy sessions.

### COD-144 — Flush queued SSE output on empty buffer-load

When a new shell session starts with an empty terminal buffer, the SSE flush was gated behind a non-empty buffer check, causing the first few output lines to be silently dropped until the next poll cycle. Removed the gate; queued output now flushes immediately on an empty-buffer load.

## Test plan

- `test/connection-indicator.test.ts` (new, 24 tests)
- `test/input-send-order.test.ts` (new, 7 tests)
- `test/ws-reconnect-plan.test.ts` (new, 6 tests)
- `test/ws-connection-registry.test.ts` (new, 6 tests)
- `test/routes/ws-routes.test.ts` (existing + 23 additive tests = 26 total)
- `test/terminal-buffer-flush.test.ts` (new, 5 tests)
- `tsc --noEmit`: 0 errors
- `npm run build`: clean
- Prettier: clean

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -483,6 +483,9 @@ class CodemanApp {
     this._clientId = '';
     this._seqCounters = new Map(); // sessionId -> last issued seq
     this._pendingDeliveries = new Map(); // sessionId -> [{seq,data,useMux,ts,tries,sentAt}]
+    // Last rendered connection-indicator tuple; the hot input path skips DOM
+    // writes when the freshly computed descriptor is identical (COD-136).
+    this._lastIndicatorDescriptor = null;
     this._postDraining = new Set(); // sessionIds with an in-flight POST drainer
     this._persistReliableTimer = null;
     this._reliableAckTimeoutMs = 4000; // unacked WS frame older than this ⇒ socket likely dead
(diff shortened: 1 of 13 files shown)


Ark0N:
Merged — thank you @aakhter! The durable-delivery hardening (per-client scoping, ACK re-drive, reconnect planning) is exactly the right direction for the exactly-once input layer. Review finished the wiring: `_wsState` now actually transitions (the chip was permanently wrong), the supersede key is per-tab (two windows on one session were 4010-evicting each other in a 5s ping-pong), backoff attempts survive the reconnect cycle (the exponential ladder was dead code), and the connected/fallback dot states got CSS. 🙙

PR #151, opened 2026-07-11
feat(image): convert HEIC paste uploads to JPEG

aakhter:
## Summary

When a browser pastes an HEIC image (common on macOS/iOS when files aren't normalised to JPEG), the paste-image route now converts it server-side to JPEG via `heic-convert` before writing to `.claude-images/`.

- Adds `convertHeicToJpeg()` helper that imports `heic-convert` dynamically, validates the output has valid JPEG magic bytes before returning
- HEIC/HEIF extensions and MIME types are now in the `ALLOWED_IMAGE_EXTS` set (they were already listed in the MIME match but the set was missing them)
- Conversion runs before the `.claude-images/` dir setup so `ext` is already `.jpg` when the filename is generated
- Returns HTTP 415 if conversion fails (unsupported media type)
- Adds `src/types/heic-convert.d.ts` type declarations
- Integration test: uploads a HEIC file, asserts the route converts it and returns a `.jpg` path

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npm run lint` — clean
- [x] `npx prettier --check` — clean
- [x] `npm test -- test/routes/session-routes.test.ts` — 54/54 passed (including new HEIC test)
- [x] `npm run build` — clean

Co-authored-by: Saqeb Akhter <saqeb.akhter@gmail.com>


Ark0N:
Merged — thank you @aakhter! HEIC→JPEG conversion removes a real iPhone paste-flow papercut. Review moved the decode off the main thread (a worker with resourceLimits + timeout — synchronous WASM decode was freezing every session's streaming for seconds per photo), capped concurrency via the existing conversion limiter, added a 64MP pre-decode guard (a <1MB crafted header could force a multi-GB allocation), and switched detection to magic bytes so mislabeled Android HEIFs convert too. 🙏

PR #152, opened 2026-07-12
feat: response-viewer (eye) support for Codex sessions

pirronewantlux529-coder:
## Problem

The response-viewer currently only reads Claude transcripts from `~/.claude/projects`. For Codex panes the eye falls back to a raw terminal-buffer dump (TUI repaint garbage), and there is no way to see the last response or the conversation thread cleanly.

A naive fix (match rollouts in `~/.codex/sessions/**` by cwd + newest mtime) breaks as soon as two Codex panes share a working directory — each pane's eye shows whatever pane wrote last. We ran that variant privately for two weeks; this PR is the design that survived real-world usage.

## Design: locating THIS pane's rollout

In order of confidence:

1. **history match** — `Session` now tracks the pane's last Enter (`codexLastSubmitAt`, both `write()` and `writeViaMux()` paths). Correlating it against `~/.codex/history.jsonl` `{session_id, ts}` entries identifies the thread the pane is *actually* on — the only signal that survives `/resume`, `/new` and `/fork` typed inside the codex TUI itself. An entry is credited to the pane whose Enter is closest among all codex panes, so a menu keystroke elsewhere can't steal attribution. Resolved pins are cached per (session, submitAt).
2. **originator match** — codex panes are spawned with `CODEX_INTERNAL_ORIGINATOR_OVERRIDE=codeman_<sessionId>`; codex (verified on 0.144.1) writes the value into `session_meta.originator` of every rollout it creates, including new files after `/new`. Newest match wins.
3. **resume-id match** — resumed rollouts keep their *original* `session_meta` (codex appends without rewriting it), so originator matching can't see them; but the rollout uuid is in the filename and `codexConfig.resumeSessionId` is known (validated as UUID first).
4. **cwd+mtime fallback** — for pre-existing panes. Hardened: case-blind path compare (codex records the launch-time case; `/mnt` paths vary on WSL), and rollouts claimed by *other* panes' originators are excluded, so an idle pane can't display a sibling's thread.

## Reader quality

- User turns come from `event_msg/user_message` — codex emits one per real user input, and injected context (AGENTS.md, `environment_context`, …) never appears there, so no fragile filtering heuristics. Legacy `response_item` user rows are deduped per-text against the event set, so mixed-version rollouts (old turns without event_msg) keep full history.
- Image inputs render an `[image ×N]` placeholder instead of silently vanishing.
- `session_meta` identity is parsed once per rollout and cached (it is write-once — verified that resume appends without touching it), so repeat eye clicks cost stats + head reads of new files only.
- `?context=full` returns the whole user/assistant thread, same shape as the Claude path.

## Frontend

- Conversation role label follows the session mode (Codex / Gemini / OpenCode / Claude) instead of hardcoding "Claude".
- The terminal-buffer fallback is now Claude-only — `_cleanTerminalBuffer` knows Claude CLI output, but for TUI modes it produced repaint garbage; those now show a clear placeholder.

## Testing

Verified on codex-cli 0.144.1, tmux backend, two machines (WSL2):

- Two codex panes created in the same directory within the same second: each eye shows its own response.
- `codex resume <id>` panes resolve via filename uuid.
- `/new` inside a resumed pane: originator match takes over (newest wins).
- `/resume` typed inside the TUI, then a new message: history match follows the pane to the resumed thread.
- Idle pane in a shared directory: empty placeholder, no bleed-over.
- Claude sessions: unchanged behavior (regression-checked).
- `tsc`, `prettier`, `npm run build` all clean.

Note: `CODEX_INTERNAL_ORIGINATOR_OVERRIDE` is an internal codex interface; if a future codex stops honoring it, layers 1/3/4 still function — the design degrades gracefully.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -2252,8 +2252,26 @@ export class Session extends EventEmitter {
    * ```
    */
   write(data: string): void {
+    this._trackCodexSubmit(data);
     if (this.ptyProcess) {
       this.ptyProcess.write(data);
+    }
+  }
+
+  // ── Codex thread tracking ─────────────────────────────────────────────
+  // When a codex pane last submitted a message (Enter). The response-viewer
+  // correlates this against ~/.codex/history.jsonl entry timestamps to find
+  // the thread the pane is ACTUALLY on — the only signal that survives
+  // /resume, /new and /fork typed inside the codex TUI itself.
+  private _codexLastSubmitAt = 0;
+
+  get codexLastSubmitAt(): number {
+    return this._codexLastSubmitAt;
+  }
+
+  private _trackCodexSubmit(data: string): void {
+    if (this.mode === 'codex' && (data.includes('\r') || data.includes('\n'))) {
+      this._codexLastSubmitAt = Date.now();
     }
   }
 
@@ -2310,6 +2328,7 @@ export class Session extends EventEmitter {
    * ```
    */
   async writeViaMux(data: string): Promise<boolean> {
+    this._trackCodexSubmit(data);
     if (this._mux && this._muxSession) {
       return this._mux.sendInput(this.id, data);
     }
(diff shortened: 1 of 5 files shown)


Ark0N:
Merged — thank you @pirronewantlux529-coder, and welcome as a first-time contributor! 🎉 The 4-layer Codex rollout resolution (history pin → originator → resume-UUID → cwd fallback) is genuinely thoughtful work, and the reviewers found the parser logic sound. Review added the missing test coverage (10 route tests over fixture rollouts covering pin/originator/uuid precedence, dedup, injected-context filtering, and a Claude-mode regression guard) plus small hardening (LRU caches, multi-block text joins). Hope to see more from you! 🙏

PR #153, opened 2026-07-12
feat(session): unified Session Manager — welcome list + searchable modal + live SSE refresh

aakhter:
Frontend companion to the unified Session Manager backend merged in #139. Where the backend added the aggregated session service + list endpoint, this adds the UI that surfaces it.

### What this adds
- **Welcome-screen session list** — the welcome overlay renders the unified session list (live + persisted + closed history, across CLIs), newest-first with richer context, rather than only Claude transcripts.
- **Persistent Session Manager modal** — an always-on header button opens a searchable modal listing every session (live + historical), with per-session mode/LIVE badges. Search filters by name, prompt, or path.
- **Live SSE refresh** — the list updates in place as sessions are created/updated/removed, no reload.
- **Session-row action menu** — each row's `⋯` opens a context menu of per-session actions.

### Scope
8 files, +502/-19 — frontend (`terminal-ui.js`, `panels-ui.js`, `app.js`, `index.html`, `styles.css`) plus a small `unified-session-service` touch and one route wire-up, and a unit test.

### Verification
- `tsc --noEmit` 0 · frontend-syntax 22/22 · prettier clean · unit tests 16/16 · build clean.
- Functional check on an isolated instance (Playwright): Session Manager button + welcome list render; modal opens and populates the unified list; live session shows with a LIVE badge; search narrows the list; the row `⋯` context menu opens with actions; creating a session live-refreshes the open list with no reload; no console errors.

Rebased onto current master (the #139 backend commits are now upstream), so this is a clean frontend-only diff.


Ark0N:
Merged — thank you @aakhter! This ties the whole session-management story together: unified welcome list with LIVE/mode badges, per-row kebab menu, and SSE-driven live refresh so the list stays current without manual reload. Because it arrived stacked on the earlier #139/#146 work that had already merged (with review fixes), we rebuilt it onto the current master rather than merging over it — kept your new pieces (the `projectKey` plumbing for 'View all in this folder', the SSE live-refresh listeners, the kebab menu and badges, the unified welcome list) while preserving master's Command Palette + fixed Session Manager (the naive merge would have re-introduced a stale duplicate that regressed the envelope error-handling and resume routing). The header button is now phone-hidden to satisfy the mobile-header policy guard. Really appreciate how much you moved Codeman forward this cycle — nine merged PRs! 🙏

PR #145, opened 2026-07-09
feat: remote host SSH cases (CRUD, case-link, session launch, injection hardening)

aakhter:
## Summary

Introduces the remote host cases feature: a domain for managing named SSH remote hosts, CRUD endpoints + case-link, and a session-UI picker that lets users start Claude sessions on a remote host via SSH. Includes the COD-107 injection-hardening fix that closes a command-injection path in `-J jumpHost` handling.

**Key files:**
- `src/remote-hosts.ts` — new domain: `RemoteHost`, `RemoteCase`, `readRemoteHosts`, `readRemoteCases`, `buildSshConnectionArgs` (fully shell-escaped), `checkRemoteTmuxAvailable`
- `src/types/session.ts` — `SessionRemote`, `RemoteCommandMode`, `defaultRemoteCommandForMode`; `SessionConfig.remote` field
- `src/web/routes/case-routes.ts` — `GET /api/remote-hosts`, `POST /api/remote-hosts`, `PUT/DELETE /api/remote-hosts/:id`, `POST /api/cases/remote-link`
- `src/web/schemas.ts` — `RemoteHostSchema`, `RemoteCaseSchema`, `RemoteCaseLinkSchema`
- `src/tmux-manager.ts` — `buildRemoteLaunchCommand` for SSH-backed sessions (ssh → remote tmux new-session); all SSH tokens via `buildSshConnectionArgs`
- `src/session.ts`, `src/web/routes/session-routes.ts` — wire `remote` through session create + quick-start; remote case resolves without local `stat` validation
- `src/web/public/session-ui.js` — remote host picker in the new-session flow
- `src/web/public/index.html` — remote host selector markup

**Security:** `buildSshConnectionArgs` shell-escapes every user-supplied field (`host`, `user`, `-p port`, `-i identityFile`, `-J jumpHost`, `-o socksProxy`, `extraSshOptions`). The `-J` flag was previously interpolated raw (command-injection); this commit closes that. A structural allowlist validates jumpHost format (`[user@]host[:port]`, multi-hop, bracketed IPv6) so no shell metacharacter can appear.

**Scope note:** This PR covers the original remote-host cases surface (COD-24) plus the injection-hardening fix. It does _not_ include durable remote tmux, discover/attach of existing sessions, or auto-reconnect — those are a follow-on.

[...]

PR #156, opened 2026-07-17
Remote tmux durability: survive SSH drop, discover/attach, collaborative sessions

aakhter:
## Summary

Adds a durability layer for remote SSH sessions, building on the remote-host SSH cases merged in #145. The agent runs inside `tmux -L codeman new-session -A` on the remote host so the session survives an SSH drop, and Codeman can discover and attach to `codeman-*` sessions another client launched there.

Scope (COD-104 to COD-109 in the downstream tracker; COD-107 already landed in #145):

- COD-104: skip the remote tmux prereq check under the test runner (test-mode).
- COD-105: discover and attach existing `codeman-*` sessions on the host. Attached (owned:false) sessions detach on tab close, never kill.
- COD-106: shared/collaborative sessions. Uses `window-size latest` so concurrent clients at different viewports do not clamp each other; adds a shared badge.
- COD-108: auto-reconnect watcher that re-attaches remote sessions after an SSH drop.
- COD-109: kill propagation for owned durable sessions.

All ssh command lines continue to flow through the single shell-safe `buildSshConnectionArgs()` introduced with #145 (the COD-107 injection hardening), so no new command-injection surface is added.

## Testing

- `tsc --noEmit`, lint, frontend-syntax, build: all clean.
- New suites: remote-discover-attach (10), remote-shared-sessions (2), remote-auto-reconnect (21), remote-kill-command (4), plus remote-discover-routes.
- Full test suite green.

Design notes: `docs/remote-sessions.md`.


Ark0N:
Merged. Thanks for the contribution! The durability layer is exactly how this should be built: the pure backoff/eligibility split in remote-reconnect.ts, the structural detach-not-kill guarantee for non-owned sessions, and keeping every ssh line routed through buildSshConnectionArgs. Great test coverage too.

Two maintainer-side additions while resolving the master conflicts (docker cases + multi-user landed in the meantime): the COD-105 discovery endpoint is now admin-gated in multi-user mode (hosts are machine-level infra, matching the hosts CRUD policy), and buildRemoteSessionCommand threads claudeMode/allowedTools through to the launch builder so the multi-user permission downgrade reaches remote agents. Also synced the doc's launch section with the shipped -L codeman-remote / codeman-ssh-<id8> naming.

PR #157, opened 2026-07-17
Session Manager: pinning, cross-device ordering, name/prompt retention

aakhter:
## Summary

Polish layer for the unified Session Manager, building on the backend service (#139) and the welcome/frontend list. Adds pinning, cross-device tab ordering, and name/prompt retention.

Scope:

- COD-131: sync tab order across devices via server state (new `src/session-order.ts`).
- COD-139: session pinning. Pinned sessions float to the top of the list.
- COD-143: retain the session name when resuming from the Session Manager.
- COD-142: retain the pin when a pinned session is killed.
- COD-140: backfill firstPrompt for sessions whose id is not the transcript UUID.
- COD-145: show the last (most recent) prompt alongside the first.

Pinning, ordering, and retention are layered onto the existing unified-session-service and welcome list. The alias-map fold that collapses a history row keyed by the conversation UUID into its owning live session is preserved.

## Testing

- `tsc --noEmit`, lint, frontend-syntax, build: all clean.
- Suites: session-order (14), session-order-routes (4), session-pin (6), session-pin-routes (5), resume-name (6), unified-session-service (26), state-store (41).
- Full test suite green.


Ark0N:
Merged. Thanks for the contribution! Pinning, cross-device tab order, and the name/prompt retention round out the Session Manager really nicely, and the pure session-order helpers plus the backfill-never-overwrites discipline made this easy to verify.

Two maintainer-side additions while resolving the master conflicts: POST /api/sessions/:id/pin now falls back to the persisted store record when there is no live session. COD-142 deliberately keeps a pinned record after kill (and cleanupStaleSessions skips pinned records), so without the fallback a pinned-then-killed session could never be unpinned. Also added modest bounds to SessionOrderUpdateSchema (id length + entry count) so a buggy client can't grow state.json unboundedly; empty strings still flow to normalizeSessionOrder as designed.

PR #162, opened 2026-07-20
fix(mobile): preserve settings across foldable postures

shenlvkang-collab:
## Summary

- keep responsive layout width-driven while giving handheld settings a stable device identity
- preserve `codeman-app-settings-mobile` when an Android foldable unfolds past the desktop breakpoint
- add an OPPO Find N5 unfolded profile and a fold → unfold → reload regression test

## Problem

Codeman chose its local settings namespace from the current viewport width. On a foldable phone, unfolding changed the viewport from phone width to desktop width and a WebView reload switched from `codeman-app-settings-mobile` to `codeman-app-settings`. Device-local opt-ins such as Response Viewer and Extended Keyboard Bar consequently disappeared.

## Fix

`MobileDetection.getDeviceType()` remains width-based for responsive layout. The settings namespace and mobile defaults now use `MobileDetection.isHandheldDevice()`, a stable touch + user-agent/client-hint classification that does not change with fold posture.

## Validation

- OPPO Find N5 unfolded viewport: 1124×1240 CSS px at DPR 2
- desktop responsive layout retained while `isHandheldDevice()` stays true
- mobile settings namespace retained after fold → unfold → reload
- Response Viewer remains visible
- Extended Keyboard Bar retains its left/right controls
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- targeted Playwright regression test
- `npm run build`

Also validated against live Codeman deployments using a fresh authenticated browser context at the unfolded Find N5 viewport.


Ark0N:
Thank you very much for testing and using it on a folding phone! Great catch: the settings namespace was chosen from the instantaneous viewport width, so unfolding past the desktop breakpoint silently switched the device to the desktop settings blob. Keeping layout width-driven while the settings identity follows the physical device is exactly the right split.

Reviewed in detail and verified locally on top of your branch:

- typecheck, lint, frontend syntax check and format check all pass
- the new OPPO Find N5 fold, unfold and reload regression test passes (chromium)
- CI is green on both jobs

One note for the record: the single failing test elsewhere in test/mobile/settings.test.ts (mobile defaults expecting subagentTrackingEnabled to be false) is pre-existing drift on master, unrelated to this change (that default was deliberately changed to true a while back and this suite is excluded from CI, so the assertion was never updated).

Merging. Thanks again!

PR #163, opened 2026-07-20
fix(paths): accept Unicode working directories

shenlvkang-collab:
## Summary

- accept Unicode letters and numbers in local working-directory paths
- keep the existing absolute-path, traversal, and shell-metacharacter protections
- cover Create Session, Quick Run, and Scheduled Run schemas with a Chinese path regression test

## Problem

`SAFE_PATH_PATTERN` only allowed ASCII letters and numbers. Valid local directories such as `/mnt/d/AI/中文项目` were rejected by `CreateSessionSchema`, `QuickRunSchema`, and `ScheduledRunSchema` before the server could check whether the directory existed. The UI consequently reported `Invalid request body` when creating or running a session from a non-ASCII working directory.

## Fix

Use Unicode property escapes for letters and numbers while retaining the existing explicit character allowlist and the `isValidWorkingDir()` checks for absolute paths, traversal, and shell metacharacters.

## Validation

- regression test accepts `/mnt/d/AI/中文项目` for Create Session, Quick Run, and Scheduled Run
- regression test still rejects the same path when a semicolon command suffix is added
- `npm test -- test/working-directory-schema.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- live API smoke test against Codeman 1.6.0: creating a shell session in a temporary Chinese directory returned HTTP 200 and preserved the exact working directory


Ark0N:
Thank you as always for the contribution! Reviewed in detail: the Unicode property escapes widen only letters and numbers, every shell metacharacter stays excluded (fullwidth homoglyphs like ； are Unicode punctuation, so still rejected), and both defense layers on top of the pattern (isValidWorkingDir in schemas.ts and isValidPath in tmux-manager.ts) are untouched. Verified the regression test, typecheck, lint and format locally; CI is green. Merging.

PR #164, opened 2026-07-20
fix(ui): show new run tabs immediately

shenlvkang-collab:
## Summary
- upsert and render newly created sessions before selecting them, without depending on `session:created` SSE timing
- fetch quick-start session details only when SSE has not already populated local state
- coalesce overlapping Run activations and keep the Run button disabled for at least 500 ms

## Validation
- `npm test -- test/run-mode-ui.test.ts` (13 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- Playwright: double-clicking Run produced one immediate tab and one `POST /api/sessions` request


Ark0N:
Thank you as always for the contribution! Nice design: keeping _onSessionCreated as the single idempotent upsert so POST-first and SSE-first orderings converge, plus the immediate tab render, is exactly the right shape, and the run() in-flight lock closes the double-click dupe. Verified locally: all 13 run-mode tests pass, typecheck/lint/frontend-syntax/public-assets/format all green, and I confirmed GET /api/sessions/:id goes through the preSerialization envelope so the fetch fallback parses correctly. One small note for a future pass: the welcome-screen buttons call runClaude()/runOpenCode()/runGemini() directly and so bypass the new lock; pre-existing behavior, not a blocker. Merging.

PR #165, opened 2026-07-20
feat(ui): add custom branding and Chinese localization

shenlvkang-collab:
## Summary

- add a server-synced custom display name under App Settings → Display
- add a per-device English / Simplified Chinese interface language picker
- localize static and dynamically inserted browser UI while preserving terminal, response, file, session-name, prompt, and path content
- apply custom branding to browser/solo-window and notification titles without renaming CLI, API, storage, or package identifiers
- validate and safely render Unicode display names

## Validation

- `npm run test:ci` — 180 files passed, 3651 tests passed (existing skips unchanged)
- focused i18n/title/auth/render suite — 46 tests passed
- frontend syntax and public-asset checks
- TypeScript, ESLint, Prettier, lockfile validation
- production build
- desktop and 393px mobile browser QA, including live switching and reload persistence

## Notes

English remains the canonical fallback. The display name syncs through server settings; language remains device-local, matching the existing display preference model.


Ark0N:
Thank you as always for the contribution! This is a thoughtfully engineered take on bolting i18n onto a plain-JS app: English-canonical keys with graceful fallback, the MutationObserver translating dynamic DOM while skipping terminal/response/file/user-content surfaces, language as a per-device display key (matching the skin pattern), and displayName strictly limited to user-facing copy with textContent/escaped-title rendering (the hostile-name test is appreciated). It also fits the project's existing direction with README.zh-CN.md and CJK input support.

Reviewed in detail and verified locally on the merged tree: the new i18n suite, typecheck, lint, frontend syntax, public assets, format, and the full test:ci sweep (3656 tests) all pass.

I updated the branch with a master merge (resolved the CLAUDE.md paragraph conflict with #162) plus two small additions in the same commit:

1. renderIndexHtml no longer recomputes windowTitle on solo-session renders. Solo renders read no settings, so the recompute reset the shared push-notification hostTitle prefix back to the default name whenever a detached window rendered.
2. Prettier-formatted test/mobile/devices.ts, which came in unformatted via the #162 merge (CI's format check only covers src/**, the pre-commit hook caught it).

Two non-blocking notes for future passes: the pre-paint language script in index.html picks the settings namespace by viewport width, which can briefly read the wrong blob on tablets and unfolded foldables until applyLocalization corrects it after load (the stable isHandheldDevice() classifier is not yet loaded at that point, so this is understandable); and the client caps displayName at 40 code points while the schema caps 40 UTF-16 units, so a name with more than 20 astral characters (e.g. many emoji) passes the client but fails the server validation.

Merging. Thanks again!

PR #170, opened 2026-07-25
feat(ui): add polished light skins

shenlvkang-collab:
Summary:\n- Add Paper Gray, Solarized Light, Catppuccin Latte, and Rose Pine Dawn skins.\n- Keep the pre-paint skin selector, native controls, mobile surfaces, xterm palettes, teammate terminals, and response-viewer states readable in light mode.\n- Refresh the local echo overlay when the terminal skin changes.\n\nVerification:\n- test/skin-themes.test.ts: 5 passed\n- typecheck, lint, frontend syntax, public asset checks, format check, lockfile check, and production build passed.\n\nThis branch is based on the current upstream master and contains only the light-skin contribution plus its changeset.


Ark0N:
Thanks for this, and welcome! This is a really solid first contribution. Rather than send you round-trips on a branch that had gone stale, I rebuilt it on current master, verified it end to end against a live instance, and pushed the conflict resolution plus two fixes straight to your branch. Everything below is what I checked and why, so you can see exactly what changed and push back on any of it.

## What you got right

You found all four places a skin has to be registered, which almost nobody gets on the first try. The token block in `styles.css`, the xterm ANSI palette in `terminal-ui.js`, the pre-paint allowlist in `index.html`, and the Settings picker. Each of those fails silently and differently on its own (a missing palette entry just leaves the terminal the wrong color, a missing allowlist entry makes the theme appear to reset on every reload), so `test/skin-themes.test.ts` guarding that parity is exactly the right test to have written.

Three details in particular stood out:

- `color-scheme: light` on the skin block. Without it the native select popups and scrollbars stay dark on a light page, and it is not obvious until you actually open one.
- `minimumContrastRatio: 4.5` for light skins, applied at construction in both `terminal-ui.js` and `panels-ui.js` and re-applied in `applyTerminalSkin()`. This is the thing that makes a light terminal actually usable, because CLIs emit colors chosen for a dark background.
- The `_localEchoOverlay?.refreshFont()` call. The overlay caches the terminal foreground and background at construction, so without that, text you have typed but not yet flushed keeps the old theme's dark backing after a live skin switch. That is a genuinely obscure interaction and you found it.

I ran a real shell session on each light skin with `ls --color`, `git status` and a raw ANSI dump. All eight colors plus bright black stay legible on every palette. Solarized Light in particular looks great.

## The bug you fixed without knowing it

Your compatibility alias block is doing more work than the comment suggests. `--bg-primary`, `--bg-secondary`, `--bg-tertiary`, `--text-primary`, `--text-secondary`, `--border-color`, `--accent-color`, `--success`, `--error`, `--danger`, `--font-mono` and `--shadow-lg` are referenced in roughly 79 places across `styles.css`, `index.html` and `settings-ui.js`, and none of them were ever declared. I confirmed this by reading computed styles on master: `--font-mono` resolves to the empty string today, so every `font-family: var(--font-mono)` in the subagent windows, orchestrator panel and file preview has silently been falling back to the inherited font this whole time. Your block makes those load-bearing. Good catch, even if it was a side effect.

## Two things I fixed on top

**1. Tinted badges are unreadable under the light skins.** A family of status badges and accent-tinted pills hardcodes pale light-on-dark ink (`#cdddff`, `#9dc0ff`, `#ffc107`, `#fff`) over a low-alpha tint. That works on a dark surface and collapses on a light one. Measured on a rendered page, twelve component families landed between 1.0 and 1.9:1. The worst case is the welcome-screen search filter chips at 1.02:1, which render as empty blue pills with no readable text at all.

Affected: search filter chips, search result badges (session, event, file), session tab mode pill, respawn state, Ralph priority and circuit-breaker badges, tunnel start button, voice provider status, history detail path and view-all button, mobile case picker items.

The fix re-points that ink at each skin's own dark tokens (`--accent-d`, `--green`, `--yellow`, `--text`) and keeps the tint as the category signal, so the color coding survives. Same components now measure 3.2 to 14:1, and the dark skins are untouched:

| component | before (light skins) | after | dark skins |
| --- | --- | --- | --- |
| search filter chip | 1.02 | 5.67 | 10.7 (unchanged) |
| search badge session | 1.37 | 5.67 | 7.96 (unchanged) |
| respawn state | 1.39 | 11.35 | 10.2 (unchanged) |
| mobile case item | 1.14 | 14.35 | 14.5 (unchanged) |
| tab mode gemini | 1.70 | 6.11 | 5.88 (unchanged) |

A few of the amber and green badges still sit at 3.2 to 4.4:1, below AA for small text. I left those alone deliberately: the dark skins run 2.1 to 2.3:1 on comparable secondary text today, so tightening it further is an app-wide design decision rather than something this PR should be on the hook for.

**2. The `--floating-bg` default repaints the existing dark skins.** Replacing the hardcoded near-black literals with a token is the right call, but the `:root` default you picked (`rgba(31,38,48,.96)`) is inherited by all three existing skins, so every modal, command palette, dropdown, subagent window, file preview and mobile sheet moved off its previous `rgba(19,19,22,.95)`.

On the Daylight skins we are keeping that. Their glass header is already `rgba(31,38,48,.85)`, so the new value is more internally consistent, and the modals now lift off the page background instead of blending into it. We compared it side by side on a live instance before deciding.

OG Codeman is pinned back to `rgba(19,19,22,.96)`. That palette is deliberately neutral near-black zinc with no blue in it, and the slate default visibly broke its identity. I added a comment on the override explaining why, so nobody removes it later.

## Housekeeping

- **Merge conflict.** Only `CLAUDE.md`, and only because master restructured it while this PR was open: deep detail moved out to `docs/architecture-invariants.md` and `CLAUDE.md` now keeps a short rule plus a pointer. Your Skins paragraph was excellent content in the wrong file after that change, so I folded the invariant into the existing "Theme skins / branding / i18n" line and expanded the detail (four-way sync, the three light-skin obligations, the shared surface tokens) into `architecture-invariants.md`. Nothing you wrote was dropped.
- **Changeset.** Expanded to cover the alias fix, the token migration and the badge fix, since it lands in the public CHANGELOG. Also dropped the em dashes to match repo style.

## Verification

Everything below is on the merged branch with the fixes:

- `npm run test:ci`: 185 files, 3762 passed, 12 skipped
- `tsc --noEmit`, `eslint`, `prettier --check`, `check:public-assets`, `check:frontend-syntax`: all clean
- CSS parses with balanced braces and no rule-count cliff (1692 to 1700 top-level rules, exactly the eight blocks added)
- Live instance in an isolated data dir and tmux socket, all seven skins, at 1440x900 desktop and 390x844 mobile, with a real PTY session for terminal output

Thanks again. This is a feature people have asked for, and the care you put into the parity test and the overlay refresh made it straightforward to land.

PR #177, opened 2026-07-28
fix(transcripts): complete tools from user results

Lint111:
## Summary

- Recognize `tool_result` blocks carried by user-role transcript entries.
- End the active tool state and emit `transcript:tool_end` with the correct tool name and error status.
- Replace fixed sleeps in watcher tests with condition-based `vi.waitFor` assertions.

## Root cause

Claude transcripts record tool requests in assistant entries but commonly record their results in user entries. The watcher only completed tools from the older result path, so Codeman could continue showing a tool as running after it had finished.

## Scope

This is the transcript completion slice extracted from #173. It changes only the transcript watcher and its focused tests. It does not alter terminal rendering, hooks, or session transport.

## Validation

- `npx vitest run --config config/vitest.config.ts test/transcript-watcher.test.ts`: 14 passed
- `npm run build`: passed
- `git diff --check`: passed

### DIFF
--- a/src/transcript-watcher.ts
+++ b/src/transcript-watcher.ts
@@ -40,6 +40,7 @@ interface TranscriptContentBlock {
   text?: string;
   name?: string;
   input?: Record<string, unknown>;
+  tool_use_id?: string;
   content?: string;
   is_error?: boolean;
 }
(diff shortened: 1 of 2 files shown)


Ark0N:
Thanks a lot! Great catch on the user-entry tool_result path: that explains tools showing as running long after they finished. Nice touch replacing the fixed sleeps with vi.waitFor too. Verified locally (14/14 tests, clean typecheck) and CI is green. Merging!

PR #178, opened 2026-07-28
fix(notifications): quiet lifecycle hook noise

Lint111:
## Summary

- Move notification preferences to schema version 5.
- Disable the drawer-only response-complete default while preserving explicit browser, audio, or push delivery choices.
- Map teammate-idle and task-completed hooks to the existing opt-in subagent categories instead of idle/stop alerts.
- Normalize local and server-hydrated preferences through the same migration path.
- Use stable handheld identity for mobile defaults and storage keys.

## Root cause

High-frequency lifecycle hooks were mapped to broadly enabled idle and stop categories, filling the notification drawer during normal agent activity. The old settings UI also reintroduced the response-complete default when preferences were hydrated from the server.

## Scope

This is the notification policy slice extracted from #173. It changes notification defaults, migration, settings persistence, and focused tests. It does not remove hook delivery or affect terminal output.

## Compatibility

The migration disables only the legacy `stop` preference shape that had no external delivery channel enabled. Users who explicitly enabled browser, audio, or push notifications keep that choice.

## Validation

- `npx vitest run --config config/vitest.config.ts test/notification-manager-noise.test.ts`: 4 passed
- `npm run build`: passed
- `git diff --check`: passed

### DIFF
--- a/src/web/public/index.html
+++ b/src/web/public/index.html
@@ -1848,7 +1848,7 @@
               <input type="checkbox" id="eventIdleAudio">
 
               <div class="event-label">Response complete</div>
-              <input type="checkbox" id="eventStopEnabled" checked>
+              <input type="checkbox" id="eventStopEnabled">
               <input type="checkbox" id="eventStopBrowser">
               <input type="checkbox" id="eventStopPush">
               <input type="checkbox" id="eventStopAudio">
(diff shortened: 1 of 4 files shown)


Ark0N:
Thank you! This is a really thoughtful cleanup: the v5 migration preserving explicit browser/audio/push choices while retiring the drawer-only stop default is exactly the right call, and routing teammate-idle/task-completed through the opt-in subagent categories kills the drawer spam without losing anything. Also glad to see the storage key aligned with the stable handheld identity (tablets and desktops are unaffected, verified). Ran the new tests, typecheck, frontend-syntax and public-assets checks locally, all green, CI green too. Merging!

PR #180, opened 2026-07-29
fix(sessions): preserve active terminal during launches

Lint111:
## Summary

- Route session-launch progress through small ownership-aware helpers.
- Keep terminal progress output on the session-less home screen.
- Use informational/error toasts when another session already owns the shared xterm.
- Apply the same behavior to Claude, shell, OpenCode, Codex, and Gemini launches.

## Root cause

Starting a new session always cleared and wrote launch status into the shared terminal. If another session was active, the later session switch could serialize that launch chrome into the active session's cached terminal frame, producing a mixed or stale view.

## Scope

This is the active-session launch-preservation slice extracted from #173. It changes only frontend launch-status routing and its focused regression test. Session creation APIs and launch sequencing are unchanged.

## Behavior

- No active session: launch progress continues to render in xterm.
- Active session: xterm is left untouched and launch status uses toasts.
- A session becoming active during launch prevents later progress lines from writing into it.

## Validation

- `npx vitest run --config config/vitest.config.ts test/run-mode-ui.test.ts`: 14 passed
- `npm run build`: passed
- `git diff --check`: passed


Ark0N:
Thank you for this contribution!

PR #184, opened 2026-07-29
fix(viewer): preserve semantic response chunks

Lint111:
## Summary

Make the response viewer preserve semantic reply boundaries and support lossless whole-chunk copy on desktop and mobile.

## Changes

- Remove terminal-buffer fallback from **Last Response**.
  - An empty structured transcript now shows the existing empty-state message.
  - It no longer presents the entire terminal history as if it were one agent reply.
- Store each reply's raw transcript text on its owning response node.
- Double-click or double-tap a reply to copy that complete raw chunk.
- Ignore copy gestures that originate from buttons, links, fields, or editable content.
- Keep existing one-click code-block copy and diagram wrap controls unchanged.
- Show brief visual confirmation after a successful whole-chunk copy.
- Use mobile terminal feedback when that optional control layer is present.

## Behavior

- **Last Response** remains a single structured reply.
- **More** remains the explicit path to the full conversation.
- Raw Markdown, whitespace, and code fences survive whole-reply copy without depending on rendered HTML.

## Validation

- `npx vitest run --config config/vitest.config.ts test/response-viewer.test.ts`
  - 2 tests passed
  - 26 frontend files passed
- `npm run build`
  - passed

## Scope

This PR does not add click-to-collapse for response messages. That interaction was discussed but is not implemented in the audited change set and should be reviewed independently from copy/source correctness.

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -1840,6 +1840,9 @@ class CodemanApp {
   _bindResponseViewerInteractions(body) {
     if (!body || body.dataset.rvBound === '1') return;
     body.dataset.rvBound = '1';
+    body.addEventListener('dblclick', (ev) => {
+      void this._copyResponseViewerChunk(body, ev);
+    });
     body.addEventListener('click', async (ev) => {
       // One-click copy: lift the raw source from the sibling <pre><code>.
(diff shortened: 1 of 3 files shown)


Ark0N:

PR #186, opened 2026-07-29
fix(mobile): route terminal taps without opening the keyboard

Lint111:
## Summary

Make terminal touches behave predictably on mobile without opening the software keyboard for TUI-owned content.

This PR is the focused tap/gesture slice extracted from the larger mobile-experience branch. It is intentionally limited to `terminal-ui.js` and its unit/browser regressions.

## Behavior

- Treat the first tap that opens the keyboard as focus-only, so it cannot also activate the currently highlighted CLI option.
- Distinguish editable prompt rows from terminal content such as readbacks, tool results, permission choices, and status rows.
- Forward content taps to Claude/Codex/Gemini as SGR mouse reports while keeping the keyboard closed.
- Keep visible prompt taps focused on xterm's helper textarea.
- Keep local scrollback taps inert while the user is reading history.
- Recognize wrapped Claude `Working` / background-agent status rows as TUI content.
- Forward touch drags to Claude's own transcript on verified Claude versions while retaining xterm-local scrolling for Codex and other modes.
- Keep the live cursor and a small lower-screen input band focusable when a Claude redraw temporarily omits its prompt glyph.

## Audit Fixes

The original classifier relied too heavily on visible prompt glyphs. During promptless Claude redraws, a live row could be classified as content, `touchstart` would suppress browser focus, and the mobile keyboard became unreachable.

The final commit gives precedence to known menus and working rows, then treats the live cursor and a four-row lower-screen fallback band as input. The browser regression verifies both required outcomes:

- xterm's helper textarea becomes `document.activeElement`
- no SGR mouse report is sent

The split also removes a hidden dependency on the separate history-anchor feature by using the existing local-scroll operations directly.

## Commit Structure

1. `fix(mobile): keep keyboard focus taps non-activating`
2. `fix(mobile): route terminal content taps to the CLI`
3. `fix(mobile): route Claude terminal gestures`
4. `fix(mobile): keep promptless terminal input focusable`

## Verification

- `npx vitest run --config config/vitest.config.ts test/terminal-touch-tap.test.ts`
  - 25 passed
- `npx vitest run --config test/mobile/vitest.config.ts test/mobile/keyboard.test.ts -t "focuses the live Claude cursor|prevents Claude subagent status taps|focuses the terminal helper textarea when the visible prompt is tapped"`
  - 3 passed, 32 skipped
- Additional focused mobile browser checks:
  - readback collapse
  - prompt focus
  - local touch scroll
  - Claude transcript drag
- `npx prettier --check test/mobile/keyboard.test.ts test/terminal-touch-tap.test.ts`
- `npm run build`

## Non-Goals

- Mobile virtual navigation controls and hardware volume-key fallback
- Centralized IME, paste, draft, or delivery state
- PTY viewport sizing ownership
- Terminal history streaming, compression, caching, or frame reconciliation

### DIFF
--- a/src/web/public/terminal-ui.js
+++ b/src/web/public/terminal-ui.js
@@ -445,7 +445,7 @@ Object.assign(CodemanApp.prototype, {
 
       let didScroll = false; // track whether touchmove fired (tap vs scroll)
       let touchStartY = 0;
-      let tapCanActivateTerminal = false;
+      let tapStartedWithTerminalFocus = false;
       const TAP_THRESHOLD = 8; // px — ignore micro-drift to distinguish tap from scroll
       container.addEventListener(
         'touchstart',
@@ -457,7 +457,7 @@ Object.assign(CodemanApp.prototype, {
             pixelAccum = 0;
             isTouching = true;
             didScroll = false;
-            tapCanActivateTerminal = this._shouldForwardMobileTapToApp();
+            tapStartedWithTerminalFocus = this._isMobileTerminalInputFocused();
             lastTime = 0;
             if (scrollFrame) {
               cancelAnimationFrame(scrollFrame);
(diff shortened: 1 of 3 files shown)


Ark0N:

PR #190, opened 2026-07-29
fix(terminal): drain deferred output without a wake event

Lint111:
## Summary

Fixes terminal output that could remain queued indefinitely after a large final response.

`flushPendingWrites()` writes one mode-specific frame budget and requeues the remainder. Previously the scheduled flag stayed set while that function ran, so its attempt to schedule the next yield was ignored. The remainder appeared only when unrelated output arrived later, which looked like truncated agent responses or passive shell commands waking only after user activity.

## Change

- centralizes terminal flush scheduling in `_scheduleTerminalWriteFlush()`
- clears `writeFrameScheduled` immediately before a flush runs
- lets a large flush enqueue its own successor until `pendingWrites` is empty
- reuses the same scheduler after flicker-buffer release
- updates the anti-flicker documentation to match xterm.js 6 and the current mode-aware frame budgets

## Validation

- `test/terminal-flush-budget.test.ts`: 7 tests passed
- new regression drains a 96 KB final Codex batch as three 32 KB writes without another terminal event
- `npm run typecheck`
- `npm run build`
- Prettier and staged diff checks

## Scope

This PR does not change terminal streaming protocols, input handling, session lifecycle, viewport ownership, or mobile controls.

### DIFF
--- a/docs/terminal-anti-flicker.md
+++ b/docs/terminal-anti-flicker.md
@@ -43,38 +43,22 @@ const syncData = DEC_SYNC_START + data + DEC_SYNC_END;
 this.broadcast('session:terminal', { id: sessionId, data: syncData });
 ```
 
-## Client-Side Implementation (`app.js`)
+## Client-Side Implementation (`terminal-ui.js`)
 
 ### `batchTerminalWrite(data)`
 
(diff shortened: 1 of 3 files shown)


Ark0N:

PR #191, opened 2026-07-29
fix(terminal): let the active viewport reclaim PTY sizing

Lint111:
## Summary

Make shared terminal sizing follow the viewport that is actively being used, without letting background restores or reconnects steal the PTY size.

This fixes the mobile-to-desktop recovery failure where the desktop canvas could remain constrained to a prior phone or keyboard layout until the user manually forced a redraw.

## Root cause

A Codeman session has one PTY size even when several browser viewports are connected. The previous arbitration treated a desktop WebSocket connection as active simply because it existed, while the browser could also send automatic resizes during restore and tab loading. That made presence, layout announcement, and real user activity indistinguishable.

The browser also repeated `fitAddon.fit()` during the same keyboard or tab transition, exposing intermediate xterm layouts and adding avoidable visual stutter.

## What changed

- Add an explicit `takeControl` resize option to the session API and WebSocket protocol.
- Keep desktop connection presence separate from recent desktop activity.
- Remember each WebSocket connection's announced viewport and dimensions, then reassert that connection's size before accepted input.
- Let explicit mobile interaction override a fresh desktop claim, and let explicit desktop interaction reclaim the PTY afterward.
- Keep automatic startup, reconnect, solo-window, and session restoration resizes passive.
- Reclaim desktop sizing on terminal/tab interaction and browser focus or visibility recovery.
- Clear stale mobile keyboard layout state before a desktop tab measurement.
- Route keyboard open/close sizing through the same resize API.
- Avoid duplicate xterm fits when the caller already established the current dimensions.

## Commit structure

1. `feat(terminal): add explicit viewport sizing ownership`
   Adds the server arbitration contract, HTTP/WS protocol support, per-socket dimension reassertion, and focused route/unit coverage.
2. `fix(terminal): let active viewport reclaim PTY sizing`
   Adds browser ownership signals, keyboard-aware layout recovery, reduced redundant fitting, and isolated Playwright coverage.

## Validation

- `npx vitest run --config config/vitest.config.ts test/session-resize-arbitration.test.ts test/routes/session-routes.test.ts test/routes/ws-routes.test.ts` - 126 passed
- `npx vitest run --config config/vitest.config.ts test/terminal-touch-tap.test.ts` - 20 passed
- Focused `test/mobile/keyboard.test.ts` viewport-control case - 1 passed
- `npx vitest run --config config/vitest.config.ts test/terminal-viewport-ownership.test.ts` - 3 passed
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- Targeted Prettier checks and `git diff --check`

The complete mobile keyboard file is not claimed green: it currently has four unrelated baseline failures covering the existing accessory list, toolbar transform, CJK/local echo, and SGR tap behavior. The new viewport-control case passes independently.

## Scope

This PR is limited to shared PTY viewport ownership and layout recovery. It does not include mobile navigation controls, terminal input persistence, bounded history streaming, response viewing, repository browsing, or session lifecycle caching.

### DIFF
--- a/config/vitest.ci.config.ts
+++ b/config/vitest.ci.config.ts
@@ -23,6 +23,7 @@ export default defineConfig({
       'test/perf-*.test.ts', // timing-sensitive perf benchmarks (flaky in CI)
       'test/inline-rename.test.ts', // browser (Playwright)
       'test/opencode-resize.test.ts', // browser (Playwright)
+      'test/terminal-viewport-ownership.test.ts', // browser (Playwright)
       'test/webgl-fallback.test.ts', // browser (Playwright)
     ],
     setupFiles: ['./test/setup.ts'],
--- a/docs/archive/phase7-test-infrastructure-plan.md
+++ b/docs/archive/phase7-test-infrastructure-plan.md
@@ -120,7 +120,7 @@ This avoids port conflicts entirely and runs fast. Only tests that need SSE or W
 | 3223 | `test/routes/ralph-routes.test.ts` | Ralph API |
 | 3224–3229 | Reserved | Future route tests |
 
-Most tests should NOT need real ports — `app.inject()` is preferred. Verified: ports 3220–3229 are completely unused by existing tests (highest used port is 3211 in `opencode-resize.test.ts`).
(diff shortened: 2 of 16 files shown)


Ark0N:

PR #192, opened 2026-07-29
fix(sessions): restore provider and working directory from live panes

Lint111:
## Summary

- persist each session provider as a tmux user option when a pane is created or respawned
- recover provider and working-directory metadata from live panes during reconciliation
- migrate legacy panes by inferring their provider once from the live process command
- preserve explicit remote and Docker workspace metadata

## Root cause

When `mux-sessions.json` was missing or contained an older recovered placeholder, reconciliation rebuilt an untracked pane with `mode: "claude"` and `workingDir: process.cwd()`. The tmux pane itself survived, but Codeman lost the provider and repository context needed to restore the session accurately.

The provider is now durable in `@codeman-mode`, while `#{pane_current_path}` supplies the live local working directory. Existing panes without the option are inferred from their process command and migrated back into tmux metadata. Stale `restored-*` paths are repaired even when Codeman restarts from a different server directory.

## Safety and compatibility

- provider values are constrained to the existing `SessionMode` allowlist
- discovered mux names and pane paths continue through existing validation
- process IDs are parsed as numbers before the `ps` fallback
- remote and Docker sessions never replace their modeled workspace with a local pane path
- legacy pid-only and path-bearing list output remains parseable

## Commit structure

1. `fix(sessions): restore provider and working directory from live panes`

## Validation

- `npm test -- --run test/tmux-manager.test.ts` (59 tests)
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npx prettier --check src/tmux-manager.ts test/tmux-manager.test.ts`
- `git diff --check`

## Scope

This PR does not include working-directory mutation UI, repository-viewer changes, terminal history streaming, or browser terminal lifecycle caching.

### DIFF
--- a/src/tmux-manager.ts
+++ b/src/tmux-manager.ts
@@ -187,8 +187,100 @@ const SAFE_TMUX_SOCKET_PATTERN = /^[a-zA-Z0-9_.-]+$/;
  */
 const PANE_LIST_SEP = '|';
 
-/** Format string for `tmux list-panes -F`. Keep in sync with {@link parsePaneList}. */
-const PANE_LIST_FORMAT = `#{session_name}${PANE_LIST_SEP}#{pane_pid}`;
+/** Format string for `tmux list-panes -F`. Keep in sync with {@link parsePaneDiscoveryList}. */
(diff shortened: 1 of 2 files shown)


Ark0N:

PR #193, opened 2026-07-29
fix(terminal): preserve replay and history ownership

Lint111:
## Summary

- keep session selection in control of its buffer-load transaction across cached and canonical replay
- keep manually scrolled terminal history anchored until the reader actually returns to the bottom

## Root causes

`chunkedTerminalWrite()` always finished the buffer-load gate, even when `selectSession()` supplied and still owned that gate. Live terminal output could therefore resume between restoration stages and interleave with a later replay.

Manual scroll protection also expired after 1.5 seconds. A busy Codex status redraw could reclaim the viewport while the user was still reading older output.

## What changed

- nested replay uses the caller's owner token without beginning or finishing that transaction
- standalone replay retains its existing self-owned behavior
- upward wheel/touch scrolling locks the viewport
- writes preserve the exact xterm viewport while that lock is active
- downward scrolling releases the lock only after xterm reports the bottom

## Commit structure

1. `fix(terminal): retain session replay ownership`
2. `fix(terminal): keep manual history scroll anchored`

## Validation

- `npx vitest run --config config/vitest.config.ts test/terminal-flush-budget.test.ts test/terminal-buffer-flush.test.ts test/codex-snapshot-replay.test.ts` (19 tests)
- `npm run typecheck`
- `npm run lint`
- `npm run build`

## Scope

This PR does not include terminal streaming, warm-session caching, viewport ownership, keyboard layout settling, or authoritative frame capture.

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -4359,9 +4359,9 @@ class CodemanApp {
         bufferWasEmpty = true;
       }
 
-      // Buffer load complete — unblock live SSE writes. chunkedTerminalWrite calls
-      // _finishBufferLoad internally (discarding queued events to prevent duplicate
-      // content); if we skipped the write (cache hit or empty), call it here.
+      // Buffer load complete — unblock live SSE writes. selectSession owns this
+      // gate across every cache/fetch/replay stage, so its chunked writes leave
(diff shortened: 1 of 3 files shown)


Ark0N:

PR #194, opened 2026-07-29
fix(mobile): coalesce keyboard viewport settling

Lint111:
## Summary

Coalesce the burst of mobile `visualViewport` resize events into one final xterm layout and PTY resize after the software-keyboard animation settles.

## Root cause

Keyboard open and close used separate fixed 150 ms and 100 ms timers. Mobile browsers report several intermediate viewport heights during the OS animation, so those timers could fit xterm against a transient height and expose another layout/redraw before the final size arrived.

## What changed

- reset one 80 ms settle timer on every viewport resize
- preserve the open/close request to return the terminal to the bottom across reschedules
- run the existing fit, padding correction, helper synchronization, and resize transport only after the viewport becomes quiet
- cancel pending settlement during handler cleanup

## Commit structure

1. `fix(mobile): coalesce keyboard viewport settling`

## Validation

- focused browser regression: 1 passed
- full `test/mobile/keyboard.test.ts`: 28 passed, 4 existing unrelated failures
- `npm run typecheck`
- `npm run lint`
- `npm run build`

The four existing failures are the stale toolbar-transform expectation, outdated accessory action list, and two local-input assertions. The new coalescing regression passes in that same run.

## Scope

This PR does not change resize arbitration, mobile terminal controls, input handling, session replay, or authoritative frame capture.

### DIFF
--- a/src/web/public/mobile-handlers.js
+++ b/src/web/public/mobile-handlers.js
@@ -209,9 +209,12 @@ const MobileDetection = {
  * Also handles terminal scrolling and toolbar repositioning via visualViewport API.
  */
 const KeyboardHandler = {
+  VIEWPORT_SETTLE_MS: 80,
   lastViewportHeight: 0,
   keyboardVisible: false,
   initialViewportHeight: 0,
+  _viewportSettleTimer: null,
+  _settleScrollToBottom: false,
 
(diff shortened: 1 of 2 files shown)


Ark0N:

PR #196, opened 2026-07-29
fix: preserve active terminal across reconnects

Lint111:
## Summary

Preserve the active terminal across same-process SSE reconnects instead of clearing state and replaying terminal history.

Extracted from #173 as a focused lifecycle fix.

## Behavior

- Add a stable `serverStartedAt` process epoch to both SSE init and `/api/status`.
- Treat an initial load, same-process reconnect, and changed server process as distinct cases.
- Reconcile authoritative session, cost, timer, subagent, and workflow metadata in place on reconnect.
- Preserve the active xterm instance, buffer, geometry, scroll position, input, and terminal caches.
- Persist available draft/input state and reload once when the server process changes.
- Ignore stale HTTP fallback responses and equal/older init snapshots.
- Drop terminal frames after a replacement page has been requested.
- Keep repository-viewer scope synchronized when a reconnect reveals a changed active working directory.

## Commits

1. `feat(server): identify init snapshots by process`
2. `fix(client): preserve terminal across reconnects`

## Verification

- `npx vitest run test/server-init-lifecycle.test.ts` - 7 tests passed
- `npx vitest run test/sse-events.test.ts` - 7 tests passed
- `npm run check:lockfile`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- Playwright at a 390x844 touch viewport against the production bundle:
  - one newer snapshot reconciled in place
  - an immediate duplicate snapshot was ignored
  - zero app resets and zero session reselections
  - active session, xterm instance, buffer object, and terminal geometry remained identical

## Scope

No terminal rendering, input parsing, gesture controls, shutdown behavior, or repository-browser implementation is included. Optional draft and repository-viewer hooks compose with their separate PRs without making them dependencies.

### DIFF
--- a/src/web/public/app.js
+++ b/src/web/public/app.js
@@ -494,6 +494,9 @@ class CodemanApp {
 
     this._initGeneration = 0;     // dedup concurrent handleInit calls
     this._initFallbackTimer = null; // fallback timer if SSE init doesn't arrive
+    this._serverStartedAt = null; // process epoch from init/status snapshots
+    this._lastInitSnapshotTimestamp = null;
+    this._serverReloadRequested = false;
     this._selectGeneration = 0;   // cancel stale selectSession loads
     this._initialFullBufferLoad = true; // first buffer load after a page load fetches full tmux scrollback (COD-47)
     this.terminalLoadStates = new Map(); // Map<sessionId, { generation, phase }>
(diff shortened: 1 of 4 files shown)


Ark0N:

PR #197, opened 2026-07-29
fix(mobile): reconcile authoritative terminal frames

Lint111:
## Summary

- Preserve the last painted xterm rows while a phone keyboard or session transition resizes the terminal.
- Reconcile Codex with a bounded 128KB `mux-visible` pane after keyboard resize, WebSocket attachment, or foreground dialogue submission.
- Hold live SSE/WS output during replacement and replay only bytes after the snapshot cursor, including cursor-crossing synchronized updates.
- Coalesce overlapping transitions, abort superseded work, and bound a stalled snapshot request to five seconds.
- Re-anchor pending local input after the authoritative pane paints without clearing existing xterm scrollback.

## Motivation

Codex can emit several resize redraws while the mobile visual viewport changes. xterm may briefly expose an older pane, blank rows, or overlapping history before the TUI reaches its final size. A visual cover removes the intermediate paint, while cursor-qualified reconciliation makes the frame behind that cover current before it is revealed.

Dialogue selection uses the same path because a prompt can be replaced immediately after Enter or a terminal mouse selection. Input is still dispatched once; the frame capture is asynchronous and cannot block delivery.

## Commit Scope

This branch is a focused terminal stack:

1. `92dbe64` retains buffer-load ownership during replay (also carried by #193).
2. `45f425d` coalesces keyboard viewport settling (patch-equivalent to #194).
3. `cc37643` drains deferred terminal output without a later wake event (patch-equivalent to #190).
4. `7bfee75` adds the inert resize-frame cover.
5. `fddd53a` adds authoritative cursor reconciliation and its tests.

It does not include the mobile navigation pad, terminal tap policy, repository viewer, shutdown endpoint, input controller, or other features from #173.

## Dependencies

Merge the terminal contracts before this PR:

1. #189 supplies bounded streamed pane snapshots plus SSE/WS cursor metadata.
2. #191 supplies explicit viewport ownership and the `takeControl` / `refit` resize options.
3. #190, #193, and #194 supply the prerequisite client commits already visible in this branch.

An integration rehearsal combined #189, #191, and this branch. Two overlapping merge blocks are additive:

- Keep stream handoff (`t: "h"`) and viewport takeover (`a: true`) in `ws-routes.ts`, with one shared `announcedViewport` declaration beside announced rows/columns.
- Keep `_requestTerminalFrameReconcile()` in `mobile-handlers.js` instead of the older direct `sendResize()` call.

The reconciled tree built successfully and passed the joint route and browser suites.

## Failure Behavior

- Runs only for touch-device, non-shell terminal sessions.
- Publishes a replacement only when the response is `mux-visible`, non-empty, and has a valid stream cursor.
- A stale session, superseding request, fetch failure, timeout, or cancellation releases the buffer gate and flushes queued live output.
- The cover is pointer-transparent and has bounded expiry; dialogue controls and terminal input remain interactive.

## Verification

- `npm run build`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- Focused unit suites: 25 passed
- Focused Chromium suites: 4 passed
- Dependency rehearsal: 158 server/client contract tests, 3 viewport browser tests, and 3 reconciliation browser tests passed
- Isolated `npm run test:ci` with #175 fixture safeguards: 189 files passed, 1 skipped; 3,817 tests passed, 12 skipped

## Related

Extracted from the terminal transition portion of #173 in response to the maintainer's scope review.

### DIFF
--- a/docs/terminal-anti-flicker.md
+++ b/docs/terminal-anti-flicker.md
@@ -73,6 +73,21 @@ this.broadcast('session:terminal', { id: sessionId, data: syncData });
 - Parallelizes session attach with buffer fetch
 - Fire-and-forget resize (doesn't block tab switch)
 
+## Mobile Keyboard Resize Cover
+
+Phone keyboard animation changes xterm's row count before the foreground TUI
+finishes its `SIGWINCH` redraw. During that interval the DOM renderer can expose
+blank or partially reflowed rows even though terminal state is valid.
+
+`KeyboardHandler` keeps the last painted `.xterm-rows` in an inert
+`.terminal-resize-frame-cover` while the viewport settles. The clone is only a
+visual cover: xterm remains the authoritative buffer, pointer events pass
+through it, and shell sessions do not use it. The cover is armed immediately
+before the resize and removed after xterm's parser callback, a short Codex
+redraw quiet window, and two animation frames. A bounded expiry removes it when
+no replacement output arrives; an active terminal buffer load keeps it in place
+until that load finishes.
+
 ## Optional Flicker Filter
 
 Per-session toggle via Session Settings. Adds ~50ms latency but eliminates remaining flicker on problematic terminals.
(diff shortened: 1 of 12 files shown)


Ark0N:

PR #200, opened 2026-08-02
fix(welcome): gate CLI welcome buttons on actual availability

timkjr:
## Summary
- Drops the always-visible Cloudflare Tunnel welcome button/QR widget — offering it regardless of whether `cloudflared` is installed is a bad default.
- Gates the Claude, Opencode, and Gemini welcome-screen buttons behind a real availability check (`/api/claude/status`, `/api/opencode/status`, `/api/gemini/status`) instead of showing them unconditionally and letting the click fail.
- Adds `isClaudeAvailable()`/`GET /api/claude/status`, mirroring the existing opencode/codex/gemini CLI resolvers — Claude previously had no availability check at all.
- Refactors the availability-check JS into a shared `_loadCliAvailability(buttonId, statusUrl)` helper instead of duplicating the fetch/try-catch per tool.

Run-mode dropdown entries (Opencode/Codex) are intentionally left unconditional here — same class of issue, but a separate follow-up to keep this PR small.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean on touched `.ts` files
- [x] `check:frontend-syntax` and `check:public-assets` pass
- [x] Verified `/api/claude/status`, `/api/opencode/status`, `/api/gemini/status` via curl against a throwaway dev server
- [x] Verified welcome-screen button visibility via Playwright: Claude/Opencode buttons visible (installed locally), Gemini button hidden (not installed), no console errors

### DIFF
--- a/src/utils/claude-cli-resolver.ts
+++ b/src/utils/claude-cli-resolver.ts
@@ -27,6 +27,15 @@ const CLAUDE_SEARCH_DIRS = [
 let _claudeDir: string | null = null;
 
 /**
+ * Returns true if the Claude CLI binary can be located (via `which` or one of
+ * the common install directories). Mirrors `isGeminiAvailable`/`isOpenCodeAvailable`/
+ * `isCodexAvailable` in the sibling resolvers.
+ */
+export function isClaudeAvailable(): boolean {
+  return findClaudeDir() !== null;
+}
+
+/**
  * Finds the directory containing the `claude` binary.
  * Checks `which claude` first, then falls back to common install locations.
  * Result is cached for subsequent calls.
--- a/src/web/public/index.html
+++ b/src/web/public/index.html
@@ -311,11 +311,11 @@
           <h1 class="welcome-title">Codeman</h1>
           <p class="welcome-desc">Manage AI Coding tools in persistent tmux sessions.</p>
           <div class="welcome-actions">
-            <button class="welcome-btn welcome-btn-claude" onclick="app.setRunMode('claude'); app.runClaude()">
+            <button class="welcome-btn welcome-btn-claude" id="welcomeClaudeBtn" style="display: none;" onclick="app.setRunMode('claude'); app.runClaude()">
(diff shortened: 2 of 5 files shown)


Ark0N:
Merged, thank you! Gating these is right, and adding `isClaudeAvailable()` / `GET /api/claude/status` closes a genuine gap: Claude was the one CLI with no availability check at all, despite being the default. The shared `_loadCliAvailability` helper was the right instinct too.

Follow-up in `5d28999` with three changes.

**The Cloudflare Tunnel button and QR widget are back, gated instead of deleted.** Your rationale is right, offering a tunnel where `cloudflared` is not installed is a bad default, but the conclusion overshoots: the welcome QR is the entire scan-to-connect-from-your-phone flow, and removing the markup left a large block of live tunnel code in `settings-ui.js` (`_updateWelcomeTunnelBtn`, `toggleWelcomeQrSize`, the SSE QR-rotation handlers) driving elements that no longer existed. Gating it on `cloudflared` is what the stated rationale actually asks for, and it keeps the feature for people who have it. New `cloudflared-resolver.ts` mirrors the CLI resolvers, and `TunnelManager` now shares its search path so the button and the spawn cannot disagree about where cloudflared lives.

**Availability now comes from an injected object rather than a fetch per surface.** Master already had this pattern from the Codex settings tab (`816d900`), and two mechanisms for one question is one too many. The status routes buy nothing as a *gating* source: every resolver memoizes its PATH probe server-side, so a fetch is exactly as stale as an injected value, while costing a round trip and leaving the buttons to flicker in after paint. The routes all stay, yours included, they are a fine API surface.

**Unknown availability now reads as available.** The `catch` hid the button on a failed fetch, so a transient blip left a working install with no run buttons at all, whereas a genuinely missing CLI only ever produced an error toast. Failing open is the cheaper mistake here. (The Codex settings *tab* keeps the opposite default, since hiding a tab costs a user nothing.)

Verified with Playwright against a real server: on this host Gemini is hidden and Claude/OpenCode/tunnel show, matching what is actually installed.

Thanks again, both of these were worth doing.

PR #201, opened 2026-08-02
fix(run-mode): gate dropdown entries on CLI availability

timkjr:
## Summary
Follow-up to #200 (welcome-screen buttons). The run-mode dropdown (gear icon next to Run) has the same problem: Claude/Opencode/Codex/Gemini entries are always shown regardless of whether the CLI is actually installed, so picking one can spawn a session that immediately errors out.

- Adds `_refreshRunModeAvailability()` in `session-ui.js`, called each time the dropdown opens; hides an entry whose `/api/<cli>/status` reports `available: false`.
- Shell is intentionally left unconditional — no external CLI dependency.

**Depends on #200**: `isClaudeAvailable()` / `GET /api/claude/status` don't exist on `master` yet, so this branch duplicates that small addition from #200 to stay self-contained and independently reviewable/mergeable. Once #200 lands, this branch should be rebased onto `master`, which will cleanly collapse the duplicate (trivial conflict, or none if git dedupes it).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean on touched `.ts` files
- [x] `check:frontend-syntax` and `check:public-assets` pass
- [x] Verified all four `/api/*/status` endpoints via curl against a throwaway dev server
- [x] Verified dropdown gating via Playwright: claude/opencode/shell visible, codex/gemini hidden (matches local install state), no console errors

### DIFF
--- a/src/utils/claude-cli-resolver.ts
+++ b/src/utils/claude-cli-resolver.ts
@@ -27,6 +27,15 @@ const CLAUDE_SEARCH_DIRS = [
 let _claudeDir: string | null = null;
 
 /**
+ * Returns true if the Claude CLI binary can be located (via `which` or one of
+ * the common install directories). Mirrors `isGeminiAvailable`/`isOpenCodeAvailable`/
+ * `isCodexAvailable` in the sibling resolvers.
+ */
+export function isClaudeAvailable(): boolean {
+  return findClaudeDir() !== null;
+}
+
+/**
  * Finds the directory containing the `claude` binary.
  * Checks `which claude` first, then falls back to common install locations.
  * Result is cached for subsequent calls.
--- a/src/web/public/session-ui.js
+++ b/src/web/public/session-ui.js
@@ -441,6 +441,7 @@ Object.assign(CodemanApp.prototype, {
     // Load history sessions when menu opens
     if (menu.classList.contains('active')) {
       this._loadRunModeHistory();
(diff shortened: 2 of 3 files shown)


Ark0N:
Merged, thank you! And thanks for splitting it out of #200 with the duplication called out explicitly, git deduped the shared hunk on merge exactly as you predicted, so this cost nothing to land.

Follow-up in `5d28999` with three changes.

**Antigravity was missing.** It is a run mode like the rest and `agy` is the least likely of the five to be installed, so it is the one that most needed gating. Not your fault, it landed after this branch. Added a static test that fails if a sixth mode reaches the dropdown without being gated, so the next one cannot slip the same way.

**The query is scoped to `menu` now.** `.run-mode-option` is also the class used by the saved-dashboard rows (`webview-tabs.js`) and the history rows, so `document.querySelector('.run-mode-option[data-mode=...]')` would find whichever comes first in the DOM rather than the menu entry.

**Availability comes from the injected object rather than four fetches per open**, same change as on #200: the server-side resolvers memoize, so a fetch is no fresher than an injected value while costing four round trips every single time the gear menu opens. Also failing open on unknown now, since the `catch` hiding an entry meant a blip could leave only Shell in the menu.

Keeping Shell ungated was the right call, incidentally, it is what guarantees the menu is never empty on a box with nothing installed. Kept and pinned in a test.

Thanks again!

PR #202, opened 2026-08-02
fix: decode dotdir working directories in history session scanning

timkjr:
## Summary

`decodeProjectKey()` in `src/web/routes/session-routes.ts` (used by `scanProjectDir()` / `GET /api/history/sessions`) reconstructs a session's real working directory from Claude Code's encoded project-key directory name under `~/.claude/projects/`.

Claude Code's encoder maps **both** `/` and `.` to `-`, so a path like `/home/user/.codeman` encodes to `-home-user--codeman`. The existing decoder's candidate-join logic only ever tried joining segments with `-` or `_`, so it could never reconstruct a hidden (dot-prefixed) directory — it would fail to find a match on disk and silently fall back to bare `$HOME`.

In practice this corrupts `workingDir` for any case that lives under a dotdir (`~/.codeman` being the most obvious example, but any project nested under a hidden directory anywhere in its path hits this), for every session resumed through the history/session-picker UI. The wrong `workingDir` then propagates into `state.json` and affects anything that trusts `session.workingDir` (CLAUDE.md lookup, paste-image dir, subagent/image watchers, etc).

## Fix

Added dot-prefixed candidate generation to both the backtracking `tryDecode()` decoder and its greedy fallback loop. A leading empty string produced by `.split('-')` is the signature of a literal `.` in the original path, so that window is retried as `.<join>` (a hidden dir) when the normal join candidates don't match anything on disk.

## Test plan

- [x] `tsc --noEmit` passes
- [x] `eslint` clean on changed files
- [x] New regression test added: `GET /api/history/sessions > decodes a dotdir working directory (e.g. ~/.codeman) instead of falling back to $HOME` — verified it fails without the fix (asserts old buggy value) and passes with it
- [x] Full `test/routes/session-routes.test.ts` + `test/routes/unified-sessions-routes.test.ts` suites pass (85 tests)

### DIFF
--- a/src/web/routes/session-routes.ts
+++ b/src/web/routes/session-routes.ts
@@ -2514,6 +2514,25 @@ export function registerSessionRoutes(
           }
         }
       }
+      // The encoder maps both '/' and '.' to '-', so a literal '.' in the
+      // original path (e.g. "/home/timkjr/.codeman") collapses into an empty
+      // split segment here. Retry this window as a dotdir/dotfile: ".<join>".
+      if (segments[idx] === '' && idx + 1 < segments.length) {
+        const dotMaxLook = Math.min(idx + 1 + 4, segments.length);
+        for (let end = dotMaxLook - 1; end >= idx + 1; end--) {
+          const dotCandidates =
+            end === idx + 1
+              ? [segments[idx + 1]]
+              : [segments.slice(idx + 1, end + 1).join('-'), segments.slice(idx + 1, end + 1).join('_')];
+          for (const child of dotCandidates) {
+            const candidate = current + '/.' + child;
+            if (await isDir(candidate)) {
+              const result = await tryDecode(end + 1, candidate);
+              if (result) return result;
+            }
+          }
+        }
+      }
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


Ark0N:
Merged, thank you! Really appreciate the work here, and especially the write-up: "the encoder maps both `/` and `.` to `-`" is the kind of root cause that is easy to miss, and you pinned it with a regression test that genuinely fails without the fix. Nice one.

One follow-up landed on top (`c15b834`), for a case the fix could not reach. The empty split segment was also being offered as a directory *name*, and `isDir(current + '/' + '')` stats `current + '/'`, which always succeeds. So the empty segment matched unconditionally:

- **backtracking half**: with a non-dot sibling present (`~/.sib` next to `~/sib`), it resolved to `/home/x//sib` instead of the dotdir. Wrong directory, plus a doubled slash that then fails every string comparison against `session.workingDir`.
- **greedy half**: that loop is shortest-match-first, so the empty candidate matched on the *first* iteration and set `matched = true`, which left your dotdir branch there unreachable.

Skipping the empty candidate in both loops fixes it, and is what makes the greedy half of your fix live. The unmatched tail needed handling too, or it would re-append a bare `/` and produce the `//` path again. Added a test asserting the dotdir wins over the sibling and the result never contains `//`.

Thanks again for chasing this one down.

Issue #208, opened 2026-08-04
Plain shell PTY exits with code 1 after successful tmux creation in Docker


DodgyBadger:
## Environment

- Codeman: 1.9.8
- Container base: `node:22-bookworm-slim` (Debian Bookworm)
- Runtime user: non-root
- Shell: `/bin/bash`
- Security configuration: `cap_drop: ALL` and `no-new-privileges`
- Codex sessions work correctly in the same container and case.

## Problem

Creating a plain-shell session results in an immediate Bash `unexpected end of file` message in the UI.

Codeman successfully creates the tmux session and spawns the shell PTY, but the PTY exits with code 1:

```text
[Session] Starting shell session with: /bin/bash (with tmux)
[Session] Created mux session: codeman-f7150602
[Session] Shell PTY spawned with PID: 478
[ws] terminal open {
  sessionId: 'f7150602-98c3-481e-b7ae-6a292695e2a5',
  wsCount: 1
}
[TmuxManager] Mouse mode OFF for codeman-f7150602
[Session] Shell PTY exited with code: 1
```

Cleanup then emits `kill ESRCH` because the process has already exited:

```text
[Session] Failed to send SIGKILL to process (already terminated): Error: kill ESRCH
[Session] Failed to send SIGKILL to process group (may not exist): Error: kill ESRCH
[TmuxManager] Killing session codeman-f7150602
```

The associated WebSocket closes with code `1005` and no reason.

## Generated launch command

I temporarily instrumented Codeman to log `launchCmd`. It generated:

```text
cd "/home/codeman/codeman-cases/testcase" && ulimit -Sn 2147483646 2>/dev/null || ulimit -n 2147483646 2>/dev/null || true && export LANG=en_US.UTF-8 && export LC_ALL=en_US.UTF-8 && unset COLORTERM && export CODEMAN_MUX=1 && export CODEMAN_SESSION_ID=97f26fcc-9568-49c8-ab4e-d0cb80ffce48 && export CODEMAN_MUX_NAME=codeman-97f26fcc && export CODEMAN_API_URL=http://127.0.0.1:3000 && export CODEMAN_HOOK_SECRET_FILE="/home/codeman/.codeman/hook-secret" && $SHELL
```

## Diagnostics performed

- `/bin/bash` works directly.
- The generated launch command passes `bash -n` with exit status 0.
- Manually reproducing Codeman's tmux creation and `respawn-pane` command succeeds and reaches an interactive Bash prompt.
[...]

PR #209, opened 2026-08-04
fix(shell): launch shell tabs as an interactive login shell

timkjr:
## Summary

Follow-up to #208's fix. Shell-mode sessions now resolve to a real absolute shell path (`resolveLocalShell()` in `src/utils/shell-resolver.ts`), but the resolved shell is launched bare — no `-i`/`-l` flags.

Without those flags, the spawned shell runs as a non-interactive child of the non-interactive `bash -c` that launches the pane, so it never sources `~/.zshrc`/`~/.bashrc`. Aliases, PATH additions, and tool init (zoxide, nvm, etc.) are silently dropped — the shell tab "works" but doesn't feel like the user's actual shell.

## Fix

Append `-i -l` to the resolved shell invocation in `buildSpawnCommand`, so it launches as a proper interactive login shell.

## Test plan

- [x] `tsc --noEmit` passes
- [x] Existing `#208` regression tests in `test/shell-session-launch.test.ts` still pass unmodified (they assert "not a bare `$SHELL`" and "parses cleanly", both still true)
- [x] Added a new test asserting the launch command ends in `-i -l`
- [x] `test/tmux-manager.test.ts`, `test/claude-permission-mode.test.ts`, `test/antigravity-mode.test.ts` all pass (76 tests)

### DIFF
--- a/src/tmux-manager.ts
+++ b/src/tmux-manager.ts
@@ -780,7 +780,11 @@ export function buildSpawnCommand(options: {
   // SERVER process's env — empty in containers and system systemd units, leaving
   // the pane command ending in a dangling `&&` ("syntax error: unexpected end of
   // file", pane dead on arrival). Resolve it in Node and quote the result.
-  return shellescape(resolveLocalShell());
+  // -i -l: interactive login shell, so the resolved shell sources the user's rc
+  // files (~/.zshrc, ~/.bashrc, etc.) instead of running as a bare non-interactive
+  // child of the non-interactive `bash -c` that launches the pane — without this,
+  // aliases, PATH additions, and tool init (zoxide, nvm, etc.) are silently dropped.
+  return `${shellescape(resolveLocalShell())} -i -l`;
 }
 
 /**
--- a/test/shell-session-launch.test.ts
+++ b/test/shell-session-launch.test.ts
@@ -88,6 +88,14 @@ describe('shell-mode spawn command (issue #208)', () => {
     expect(cmd.trim()).not.toBe('');
   });
 
+  it('launches an interactive login shell, so rc files (~/.zshrc, ~/.bashrc, etc.) are sourced', () => {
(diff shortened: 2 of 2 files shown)


Ark0N:
Merged, thank you! Login-shell parity is the right call: it matches what tmux itself does for a pane with no `default-command`, and it is what picks up `/etc/profile` and `/etc/profile.d/*`, which a systemd `--user` service never sourced. That was a real gap.

Follow-up in `2b89f35` with two corrections.

**The flags are now allowlisted rather than hardcoded.** They get appended to a path that ultimately comes from the passwd entry, which is user data and can name anything. nushell, elvish and xonsh take neither `-i` nor `-l` in this form, and a shell that rejects an unknown flag exits on the spot, so a user with one of those in passwd would have gotten a dead pane on arrival, which is #208 all over again. `loginShellArgs()` applies them only to the POSIX-family shells verified to accept both, and a test really launches every allowlisted shell present on the machine rather than trusting the set. csh/tcsh are excluded deliberately: tcsh honors `-l` only when it is the *only* flag, so `-i -l` would not have been a login shell there anyway.

**On the rationale**, one correction worth recording since the comment will outlive us both: a tmux pane already hands the shell a tty, so it was interactive all along and `~/.bashrc` was always being sourced. Checked it directly:

```
$ tmux new-session -d 'bash -c "exec /bin/bash"'   # then, in the pane:
DASH=[himBHs] LOGIN=[no]
```

`$-` contains `i`. So `-l` is the flag doing the work here, not `-i`. Worth knowing because `-l` is not purely additive for bash: an interactive *login* bash reads `~/.bash_profile` and skips `~/.bashrc` unless the profile sources it, which I confirmed by loading a fake HOME whose `.bash_profile` does not. Ubuntu's stock `~/.profile` does source it, and anyone running tmux by hand already gets a login shell, so this is the right default. Updated the comments to say so.

Thanks again, good fix.

PR #210, opened 2026-08-04
fix(remote-ssh): route shell + agent CLIs through a real interactive login shell

timkjr:
## Summary

Remote SSH shell/agent-CLI sessions weren't reaching the remote user's real environment, for two separate reasons:

**1. Shell mode hardcoded `exec bash -l`**, ignoring the remote user's actual login shell. sshd sets `$SHELL` from the remote user's `/etc/passwd` entry, so this should launch their real shell (zsh, fish, etc.), not always bash.

**2. `remain-on-exit` was only ever set on the local tmux socket.** If the remote pane command exited for any reason — even something transient — tmux destroyed the pane, window, and (being the only session) the whole remote server, tearing the local `ssh -t` attach down with it and leaving no trace to diagnose. Reconnect's `-A` then created a fresh session, which could repeat as a flap loop with no evidence surviving between attempts.

**3. Once `remain-on-exit` revealed the dead panes**, the real underlying failure became visible: `exec claude`/`exec opencode`/etc. run under ssh's non-interactive, non-login remote-command shell, which only sees sshd's minimal default PATH — not the `~/.zshrc` PATH entries where these CLIs actually live (e.g. `~/.local/bin`, `~/.opencode/bin`). They were silently failing with "command not found" (exit 127).

## Fix (two commits)

1. `shell: 'exec $SHELL -i -l'` instead of hardcoded bash; add `remain-on-exit on` to the remote tmux session's `set` options.
2. Route `claude`/`opencode`/`codex`/`gemini`/`antigravity` through `$SHELL -i -l -c '<cmd>'` too, so PATH is fully resolved before the CLI name is looked up — mirroring the shell-mode fix.

## Test plan

- [x] `tsc --noEmit` passes
- [x] Updated existing assertions in `test/remote-hosts.test.ts`, `test/remote-ssh-options.test.ts`, `test/tmux-manager.test.ts`, `test/antigravity-mode.test.ts` for the new default commands (81 tests pass)
- [x] Per-host `commands.*` override fixtures (e.g. `aaDesktop`, `remote-shared-sessions.test.ts`) deliberately left untouched — they test the override escape hatch, not the default
- [x] Docker-mode default commands (`defaultDockerCommandForMode`) confirmed unaffected — separate function, separate concern

### DIFF
--- a/src/remote-hosts.ts
+++ b/src/remote-hosts.ts
@@ -60,7 +60,11 @@ export async function writeRemoteCases(configDir: string, cases: RemoteCase[]): 
 
 export function defaultRemoteCommandForMode(mode: SessionMode): string {
   const commands: Record<RemoteCommandMode, string> = {
-    shell: 'exec bash -l',
+    // $SHELL, not a hardcoded bash: sshd sets it from the remote user's
+    // /etc/passwd entry, so this launches their actual login shell (zsh,
+    // fish, etc.). -i -l so it sources rc files (~/.zshrc etc.), matching
+    // the local shell-mode launch.
+    shell: 'exec $SHELL -i -l',
     // Mirror the LOCAL claude default so the remote agent runs non-interactively
     // (no trust-folder/permission prompt that nothing on the remote answers). The
     // per-host `commands.claude` override stays the escape hatch.
--- a/src/remote-hosts.ts
+++ b/src/remote-hosts.ts
@@ -59,6 +59,15 @@ export async function writeRemoteCases(configDir: string, cases: RemoteCase[]): 
 }
 
 export function defaultRemoteCommandForMode(mode: SessionMode): string {
+  // Agent CLIs (claude/opencode/codex/gemini/antigravity) are typically installed
+  // under per-user paths like ~/.local/bin or ~/.opencode/bin, added to PATH only by
+  // the remote user's interactive-login shell startup files (~/.zshrc etc.). ssh's
+  // remote-command execution is neither interactive nor login, so a bare `exec
+  // claude` sees only sshd's minimal default PATH and fails with "command not
+  // found" (exit 127) — confirmed via `tmux capture-pane` on the
+  // remain-on-exit-preserved dead pane. Route through `$SHELL -i -l -c`, the same
+  // fix already used for shell mode below, so PATH is fully resolved before the
+  // CLI name is looked up.
   const commands: Record<RemoteCommandMode, string> = {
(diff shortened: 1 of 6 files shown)


Ark0N:
Merged, thank you, this is excellent debugging. Using `remain-on-exit` to make the dead pane survive long enough to `capture-pane` it, and finding the exit-127 PATH failure hiding underneath the flap loop, is exactly the right way to chase something that destroys its own evidence. The fix itself is solid and I verified it end to end: ran the emitted pane command through all three quoting layers under a minimal sshd-style PATH, with a stub CLI installed only on a login-shell PATH entry. It resolved and launched with arguments intact and a space-containing remote path preserved.

Follow-up in `2b89f35` with two corrections to the `remain-on-exit` line.

**`on` -> `failed`.** `on` also keeps the pane after a *clean* exit, so typing `exit` in a remote shell strands a dead pane, the session outlives it, and the next launch's `-A` reattaches to that corpse instead of starting a shell. Permanently, on the default path. Verified against a real tmux:

```
$ tmux new-session -d -s codeman-ssh-abc '...' \; set remain-on-exit on
# type `exit` in the pane, then simulate reconnect:
$ tmux new-session -A -d -s codeman-ssh-abc '...'
pane_dead=1   ->   "Pane is dead (status 0)"
```

`failed` keeps the pane only on a non-zero exit, so status 0 tears the session down as before and a 127 still leaves "command not found" on screen. That is precisely the diagnostic case you wanted, without the strand.

**Moved last in the chain.** tmux aborts the remaining commands of a `\;` sequence once one errors, also verified:

```
$ tmux set -t s bogus-option-xyz on \; set -t s status off
invalid option: bogus-option-xyz
$ tmux show -t s -v status
(empty -- the second command never ran)
```

`failed` needs tmux >= 3.2 on the *remote* host, so leading it meant a rejection there would silently drop `status`/`mouse`/`prefix`/`escape-time`/`window-size` along with it. Trailing, a rejection costs only that one option.

Also folded `$SHELL` into `"${SHELL:-/bin/sh}"` behind one shared `remoteLoginShellCommand()` helper, so the string is not rebuilt in `tmux-manager` as well. In fairness that default is belt-and-braces rather than a live bug: tmux always exports `SHELL` into a pane from its own `default-shell`, so the command as used was safe either way. It matters because those strings are the seed values a per-host `commands.*` override gets edited from.

Great work, thanks for pushing this one through to the actual root cause.

PR #214, opened 2026-08-05
fix(mobile): gate the phone overview's run picker on CLI availability

timkjr:
## Summary

The phone overview's Run picker (the "C" logo home screen under 430px) always listed all six CLI backends — Claude, Shell, OpenCode, Codex, Gemini, Antigravity — regardless of whether the corresponding CLI was actually installed. Tapping an unavailable one produced a confusing failed-launch instead of the entry simply not being there.

The desktop/tablet toolbar's Run dropdown already gates each entry on `isCliAvailable(mode)`; the mobile overview's picker in `mobile-overview.js` never had the equivalent check, so it silently drifted from the toolbar's behavior as new CLI backends were added.

## Fix

`_buildMobileOverviewRunMenu()` now skips any non-shell entry whose CLI isn't installed, mirroring the toolbar's existing gating logic exactly (same `isCliAvailable()` call, same shell exemption).

## Testing

- `test/mobile-overview.test.ts` (vm-sandbox harness) — new coverage asserting an unavailable CLI is excluded from the menu while available ones and shell remain.
- `tsc --noEmit` clean, `npm run lint` clean.

### DIFF
--- a/src/web/public/mobile-overview.js
+++ b/src/web/public/mobile-overview.js
@@ -486,6 +486,11 @@ Object.assign(CodemanApp.prototype, {
    * The Run picker: the same backends as the toolbar's run-mode menu, plus saved
    * web tabs. Deliberately no "Recent Sessions" block, unlike the toolbar menu:
    * past conversations have their own section further down this screen.
+   *
+   * Gated the same way as the toolbar's #runModeMenu (isCliAvailable(), shell
+   * exempt) — this list is a separate, hardcoded duplicate of the toolbar's menu
+   * rather than a shared render, so it never picked up #201's gating and offered
+   * every backend regardless of what's actually installed.
    */
   _buildMobileOverviewRunMenu() {
     const menu = document.createElement('div');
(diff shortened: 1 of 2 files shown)


Ark0N:
Merged, thank you! 🙏

Clean catch. The phone overview's picker being a hardcoded duplicate of the toolbar menu rather than a shared render is exactly why it silently missed #201's gating, and it would have kept drifting every time a backend was added. Mirroring the toolbar logic exactly (same `isCliAvailable()`, same shell exemption) is the right fix, and `isCliAvailable()` failing open on an unknown flag means the menu can never end up empty.

The third test is the one I appreciate most: asserting that every mode the picker *offers* is actually gated catches the next backend being added ungated, which is the recurring shape of this bug rather than the instance of it.

Verified here before merging: merged cleanly alongside #203 and #215, and the full `test:ci` sweep on the combined merge was green (198 files, 3971 tests, zero failures).

Thanks for keeping the mobile surface in step with the toolbar.

PR #1, opened 2026-01-24
feat: add HTTP Basic Auth for web interface security

ryan12324:
Add optional password protection via environment variables:
- CLAUDEMAN_PASSWORD: enables auth when set
- CLAUDEMAN_USERNAME: optional, defaults to "admin"

Uses dotenv to load .env file. When enabled, browsers automatically prompt for credentials. Backward compatible - no auth if password unset.

PR #215, opened 2026-08-05
fix(history): three Past Sessions data-quality bugs (automated-session noise, cross-contaminated previews, blank restart-heavy rows)

timkjr:
## Summary

Three data-quality bugs in the Past Sessions list (Cmd+K Session Manager / phone overview PAST SESSIONS), all rooted in `scanProjectDir()`/`extractTranscriptEntrypoint()` in `session-routes.ts` and `mergeUnifiedSessions()` in `unified-session-service.ts`. Bundled together since they touch the same history-scanning path and were found while investigating the same user-facing symptom (a noisy, sometimes-wrong session list).

## 1. Automated/SDK-driven transcripts cluttering history

Claude Code stamps an `entrypoint` field on message records: `'cli'` for a real interactive session, `'sdk-py'`/`'sdk-cli'` for an SDK-driven one (CI review bots, automated tooling, etc.). Those were showing up in Past Sessions indistinguishably from real sessions a user could actually resume into — they never were resumable, so they were pure noise.

`extractTranscriptEntrypoint()` now scans a transcript's `"type":"user"`/`"type":"assistant"` lines for the entrypoint field and excludes the transcript only when **every** entrypoint-bearing message is non-`'cli'`. On my own `~/.claude/projects` (321 transcripts as of this PR), that's **257 excluded (80%)** — 76 `sdk-cli`, 178 `sdk-py`, 3 with no entrypoint field at all (left visible, fail-open). This is a **hard exclusion, not a toggle** — deliberately: these transcripts were never something a user could resume into, so there's no legitimate "show me the automated ones too" use case in this list. Fully scoped to Claude Code transcripts only; the other four CLI backends (OpenCode, Codex, Gemini, Antigravity) have their own separate history-reading code paths, untouched by this change (confirmed by researching each one's actual on-disk session format before writing this).

The detection logic itself went through a second pass after an independent review caught a real bug: the original version returned the **first** entrypoint-bearing message's value rather than scanning all of them, so a transcript that started under an older Claude Code build (no entrypoint field on its true first message) and later picked up a non-`'cli'` entrypoint on some later message was wrongly excluded — the opposite of the intended fail-open behavior. It now returns `'cli'` the moment *any* scanned message carries it.

## 2. Cross-contaminated firstPrompt/lastPrompt on resumed sessions

`mergeUnifiedSessions()`'s workingDir-based backfill (used to recover a `firstPrompt`/`lastPrompt` for a session row whose own transcript extraction came up empty) didn't check whether the row already had its own history entry. A resumed session with its own aliased transcript could have its real preview text overwritten by a same-directory sibling's — showing the wrong conversation's content next to the right session. Fixed by skipping the backfill whenever the row already has an entry from its own `history` source; borrowing only ever applies to rows with no history entry of their own.

## 3. Blank rows from restart-bookkeeping accumulation, plus a real perf fix

Every session restart writes a small batch of bookkeeping-only lines (`mode`, `permission-mode`, `last-prompt`, `queue-operation`) ahead of the next real message. A session restarted many times over a long conversation accumulates enough of these that the original 16KB head-read window landed entirely on bookkeeping, producing a blank row.

This went through two iterations:
- First fix: raised the head-read buffer to 128KB unconditionally, matching an existing precedent elsewhere in the file. That fixed the blank rows but, per the same independent review, roughly **quadrupled per-request scan cost** — measured on my real transcript tree, ~4x both bytes read and wall time — since every file now paid the 128KB cost even though only a minority actually need it.
- Current fix: two-tier read — try 16KB first, escalate to 128KB only when that wasn't enough. Verified against my real `~/.claude/projects` tree (320 files): **36% fewer bytes read, ~17.5% faster wall time** than the unconditional-128KB version, with identical output.

Also fixes a fallback regression introduced during that same rework: a failed head read (e.g. `EMFILE` while scanning hundreds of files) on a file at or under the head-buffer size no longer got a shot at the tail-read fallback, silently dropping the session from history instead of giving it a second chance.

## Testing

- `test/routes/session-routes.test.ts` (87 tests) — includes new coverage discriminating the "any cli wins" vs "first field wins" entrypoint logic (verified to fail against the pre-fix version), and the restart-bookkeeping head-window case.
- `test/services/unified-session-service.test.ts` (28 tests) — includes new coverage for the resumed-session cross-contamination case (verified to fail against the pre-fix version).
- `tsc --noEmit` clean, `npm run lint` clean.
- All three fixes verified against real production transcript data (live `curl` requests, before/after) as well as the automated suite.

### DIFF
--- a/src/services/unified-session-service.ts
+++ b/src/services/unified-session-service.ts
@@ -226,6 +226,16 @@ export function mergeUnifiedSessions(sources: UnifiedSources): UnifiedSessionIte
   // different UUID. Backfill from the already-passed history: first try the claudeSessionId
   // join, then the newest transcript in the same workingDir. Never overwrite a non-empty
   // firstPrompt (so rows keyed to their own transcript are untouched).
+  //
+  // The workingDir guess is a last resort and MUST be skipped for any item that
+  // already has its own 'history' entry (step 1 above already gave it a real,
+  // direct scan of its own transcript). Without this guard, a history row whose
+  // OWN extraction genuinely failed (oversized first message, etc.) silently
+  // inherited the newest OTHER session's opening line from the same directory —
+  // not a blank, but actively wrong: old sessions displayed today's conversation
+  // as if it were their own. A row with no 'history' source at all (its
+  // transcript hasn't been linked/scanned under its own id yet) has no such
+  // direct attempt to prefer, so the guess remains a reasonable stand-in there.
   const firstPromptByUuid = new Map<string, string>();
   const firstPromptByWorkingDir = new Map<string, { prompt: string; ms: number }>();
   // COD-145: lastPrompt rides the same backfill (build parallel indexes; never overwrite).
@@ -254,12 +264,13 @@ export function mergeUnifiedSessions(sources: UnifiedSources): UnifiedSessionIte
     }
   }
   for (const item of map.values()) {
+    const hasOwnHistoryEntry = item.sources.includes('history');
     if (!item.firstPrompt) {
       // never overwrite an existing non-empty prompt
       const byUuid = item.claudeSessionId ? firstPromptByUuid.get(item.claudeSessionId) : undefined;
       if (byUuid) {
         item.firstPrompt = byUuid;
-      } else if (item.workingDir) {
+      } else if (item.workingDir && !hasOwnHistoryEntry) {
         const byDir = firstPromptByWorkingDir.get(item.workingDir);
         if (byDir) item.firstPrompt = byDir.prompt;
       }
@@ -268,7 +279,7 @@ export function mergeUnifiedSessions(sources: UnifiedSources): UnifiedSessionIte
       const byUuid = item.claudeSessionId ? lastPromptByUuid.get(item.claudeSessionId) : undefined;
       if (byUuid) {
         item.lastPrompt = byUuid;
-      } else if (item.workingDir) {
+      } else if (item.workingDir && !hasOwnHistoryEntry) {
         const byDir = lastPromptByWorkingDir.get(item.workingDir);
         if (byDir) item.lastPrompt = byDir.prompt;
       }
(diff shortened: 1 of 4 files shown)


Ark0N:
Merged, thank you! 🙏

Three real bugs, each with the reasoning shown, and the two-pass history on #1 and #3 (fixing the first attempt after review caught the "first field wins" misattribution and the 4x scan cost) is exactly the kind of iteration I like to see in a PR description rather than hidden in force-pushes. The two-tier 16KB→128KB read is a genuinely better answer than either extreme.

Verified here against real data before merging:
- Your five new tests fail on unmodified `master` and pass on the merge.
- Ran the merged build side by side with `master` against my real `~/.claude/projects`. Of the 36 rows present in both lists, exactly one changed, and it changed from `(no content)` to its real first prompt: fix #3, confirmed on live data.
- Every row newly excluded by fix #1 was `sdk-cli`-only, with no `cli` entrypoint anywhere in the file. Zero false positives.
- Full `test:ci` sweep on the combined merge with #203 and #214: 198 files, 3971 tests, zero failures.

**One follow-up I pushed on top** (`ecd3f3f`), flagging it since it changes your logic:

The exclusion shipped as `entrypoint !== 'cli'`, which is an allowlist on a value. Because the check *hides* rows, it fails closed on anything Claude Code has not shipped yet: if a future release renames the interactive entrypoint or adds a second interactive host, nothing matches `'cli'` and the whole Past Sessions list goes blank with nothing in the UI to explain it. I inverted it to a blocklist on the SDK shape (`sdk`, `sdk-cli`, `sdk-py`), so an unrecognized automated entrypoint costs a few noisy rows instead of a dead feature.

This is the same fail-open reasoning your own comment applies to a *missing* entrypoint field ("matching every other gating check in this codebase"); only the unknown-*value* case was inverted. Added a test that fails against the original line. No change to behavior on any entrypoint value that actually exists today, so your verification still holds.

Thanks for the thorough work on this one.

PR #216, opened 2026-08-06
fix(web): align brief Response Viewer formatting

shenlvkang-collab:
## What changed

The brief Response Viewer path rendered Markdown directly under `#responseViewerBody`, while the More/full-history path wrapped each message in `.rv-message > .rv-text`. That made headings, paragraphs, tables, and code blocks use different layout rules and could make the brief view look malformed.

This change uses one message builder for both paths, keeping the brief assistant response at the same formatting quality as the full conversation view.

## Verification

- `npm run typecheck`
- `npm run lint`
- `npm run check:lockfile`
- targeted Vitest: 14 tests passed
- production build passed
- Playwright beta smoke test: default view and More view both rendered the shared message/card structure

The change is based on official `codeman@1.11.1`.

### DIFF
--- a/.changeset/brief-response-viewer-parity.md
+++ b/.changeset/brief-response-viewer-parity.md
@@ -1,0 +1,5 @@
+---
+"aicodeman": patch
+---
+
+Keep the brief Response Viewer output inside the same message card and Markdown wrapper used by the full conversation view, so opening the viewer without clicking More preserves the same readable formatting.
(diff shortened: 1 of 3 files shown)


Ark0N:
Merged, thank you! Nice catch on the brief view. Pulling the message building into `_buildResponseViewerMessage()` is the right fix: it makes the two paths share the card, role badge and `.rv-text` markdown scoping instead of keeping them in sync by hand, so the formatting can't drift apart again.

Appreciate the thorough verification notes on the PR too, and the changeset. Shipping in 1.11.2.

Issue #154, opened 2026-07-13
Can't scroll back through terminal history on macOS trackpad (Claude sessions, 1.3.0)


Majow:
## Environment

- **Codeman:** 1.3.0 (self-hosted, git install)
- **Claude Code CLI:** 2.1.204
- **Browser:** Chrome on macOS (desktop)
- **Input device:** MacBook **trackpad** — *no external mouse tested* (see caveat below)

## Problem

In a Claude session I can no longer scroll back through the terminal output. Scrolling up with a two-finger swipe only reaches roughly **one screen (~34–36 lines)**; as new output arrives, older lines disappear and I can't scroll back to them. This used to work normally around **1.2.x** — a plain two-finger scroll reached the terminal's scrollback.

> **Caveat:** I use a MacBook **trackpad** and have **not** tested with a physical mouse wheel, so I can't confirm whether Shift+wheel behaves differently there. This report is specifically about the trackpad experience.

## Steps to reproduce

1. Open a Claude session in the web UI on macOS/Chrome with a trackpad.
2. Produce more output than fits on screen (e.g. ask Claude to print 1–50).
3. Two-finger scroll up.

**Expected:** reach the top of the output (line 1).
**Actual:** only reach ~line 34–36; earlier lines are gone and can't be scrolled back to.

## What I found while digging (may help narrow it down)

- I captured ~54 KB of the live PTY stream from a Claude session. It is entirely **full-frame repaints**: synchronized-update markers (`CSI ?2026h/l`, ~2130 of them) and absolute cursor addressing (`CSI <row>;<col>H`, ~2130), with **0 newlines** in the whole stream. So Claude Code (≥ 2.1.187) repaints the viewport in place and never scrolls output into the terminal's scrollback. `tmux` confirms the pane is in the alternate screen with `history_size=0`.
- `_shouldForwardWheelToApp` (from #144) forwards the plain wheel to the CLI for claude ≥ 2.1.187 (so Claude scrolls its own transcript), and moves the local xterm scrollback onto **Shift+wheel**.
- On a **macOS trackpad**, holding Shift while two-finger scrolling is interpreted by the browser as **horizontal** scroll (`deltaY ≈ 0`), so the `Shift+wheel` branch of the container wheel handler never fires. Combined with the point above, there doesn't seem to be a working gesture for a trackpad user to page back through a Claude session's history.

## Expected behavior

A macOS trackpad user should be able to scroll back through a Claude session's history the way it worked in ~1.2.x.

[...]

PR #223, opened 2026-08-07
fix: terminal scrollback overhaul for shell and CLI sessions (#205)

### LINKED ISSUE #205: Scrollback in terminal not working
I am starting a shell, not an LLM. Since claude seems to scroll at least somewhat.

When starting a shell and printing a bunch of lines, I should be able to scroll down. But in my android, scrolling on the term screen does nothing.

Then I switched to it on my desktop. And I can see the mouse wheel scrolls back through previous commands, like pressing up. But does not scroll the screen. This is likely a tmux setting?

I'm on Mac M4. With brew install tmux.

Ark0N:
Fixes #205.

## What was broken

Five distinct problems hid behind one issue (full evidence in `docs/scrollback-issues-analysis.md`, measured against the live 1.11.2 instance):

1. **Shell/opencode/antigravity sessions had no scrollback at all.** The tmux CLIENT emits `smcup` as its first bytes at attach, parking xterm in the alternate buffer for the whole session: touch scrolling no-ops (Android symptom) and xterm converts the wheel into Up/Down arrows, so the shell cycles command history (desktop symptom). Both reported symptoms, one sequence.
2. **Bursty output destroys browser scrollback in every mode.** tmux coalesces fast output into pane repaints instead of linefeeds: a 60-line burst added 1 row of browser history and destroyed 34 (measured), while tmux's own history stayed complete.
3. **Tab switches collapsed history to one frame.** The single `full=1` replay per page load was consumed by whichever tab auto-selected.
4. **Firefox scrolled ~4x too slowly** (`deltaMode` line units read as pixels).
5. **Remote SSH Claude sessions never got a CLI version probe**, silently disabling wheel forwarding (residual #154).

## What this changes

- **Narrow alt-screen strip** for tmux-backed shell/opencode/antigravity (`isMuxAltScreenOnlyStripMode`): strips ONLY tmux's own client smcup, keeps `3J` and mouse DECSETs. tmux never forwards a pane program's alt-screen toggles (measured: vim/less emit zero to the client), and the `useMux` gate keeps direct-PTY fallback sessions untouched so vim/less/htop stay correct there. Applied on the live path and the replay path.
- **Per-session `full=1`** (Set instead of one-shot flag) plus an **on-demand history re-pull**: scrolling up at the top of the buffer re-fetches the complete tmux scrollback (4s cooldown, in-flight and tab-switch guards, viewport position held), recovering everything repaints or tab switches ate.
- **Wheel AND touch forward to the CLI transcript** for codex / claude >= 2.1.187 at any scroll position (snap to bottom first, 40ms coalescing, tick cap). The old viewport-at-bottom gate pinned users to a buffer of stale repaint frames and let the CLI's input box scroll off screen; touch previously always scrolled locally, which was the mobile half of the bug. Shift+wheel and the `terminalWheelLocalScrollback` setting still pin local scrollback.
- **`deltaMode` normalization** in `_wheelScrollLines()` (line/page/pixel units), keeping the #154 Shift-axis trap.
- **Remote CLI version probe over ssh** (`probeRemoteCliVersion`), deferred at session start, using the same `buildSshConnectionArgs` + login-shell wrapper as the real launch.

The original diagnosis and the divergence from its recommended approach are recorded in `docs/scrollback-fix-plan.md`; invariants updated in `docs/architecture-invariants.md` and CLAUDE.md.

## Testing

- `npm run test:ci`: 198 files, 3983 tests, all passing (includes new unit tests for the strip flavors, deltaMode conversion, forwarded-scroll snap, and the ssh probe command builder).
- E2E with Playwright against a live beta instance, driven by trusted events (`page.mouse.wheel`, CDP touch), asserting on xterm buffer state and real PTY bytes:
  - Shell session (10/10): normal buffer at attach, wheel emits zero arrow keys, wheel-up at top re-pulls all 300 test lines, less enters/leaves cleanly, reload replays into the normal buffer.
  - Claude session (6/6): version probe lands, desktop wheel and touch drags in both directions forward SGR reports with the viewport pinned to the live screen.


Ark0N:
Follow-up commit c067167 after beta testing surfaced two more failure modes (thanks for the exact repro: reload, switch tab, come back; and the input box scrolling up then fighting):

Measured against the live instance, xterm's vscode-style viewport scroller consumes wheel events itself whenever it believes a scrollbar exists, does not consult attachCustomWheelEventHandler, and goes deaf after terminal.reset(): following a tab switch or full-history replay it neither scrolls nor propagates. So the bubble-phase container handler never ran once local scrollback existed (forwarding, deltaMode and the top-of-buffer re-pull were all silently dead exactly on sessions WITH history), and after a tab switch nothing scrolled at all.

The wheel listener now runs in capture phase, stops propagation, and scrolls locally through buffer-level scrollLines(), which keeps working after resets. Mouse-tracking sessions and the alternate buffer (direct-PTY vim/less) are passed through untouched.

Re-verified end to end on the beta: 9/9 matrix checks covering the reported flows (claude wheel with scrollback present stays pinned and forwards; shell reaches full history by wheel alone; reload then tab switch then back still works; SSE reconnect survives; Shift+wheel stays local), plus the two earlier E2E suites re-passing 10/10 and 6/6.

PR #224, opened 2026-08-07
fix(http): raw writeHead routes drop every header the security hook set

christianhaberl:
## The bug

`reply.raw.writeHead()` writes straight to the Node response and bypasses Fastify's header store, so everything `registerSecurityHeaders` set in its `onRequest` hook is silently dropped on every route that answers that way.

The visible symptom is CORS. The hook emits `Access-Control-Allow-Origin` for localhost origins, so a page served from a local dev server may call every `/api` endpoint cross-origin — except the four below, whose requests fail. `X-Content-Type-Options`, `X-Frame-Options` and the CSP were being lost the same way.

## Affected

| Route | File |
|---|---|
| `GET /api/events` | `src/web/server.ts` |
| `file-raw` · `tail-file` · `download` | `src/web/routes/file-routes.ts` |

The codebase already contains this exact pattern in `sendRawStream` (`file-routes.ts`), which is what suggests the other four sites are an oversight rather than a decision.

## Tests

Five tests against a real `WebServer`, including a comparison of `/api/events` against `/api/status` for the same Origin — the point being that the SSE route stops being the odd one out.

Verified in both directions: **with `...inherited` removed from the `/api/events` handler, 3 of the 5 fail.**

## Honest limitations

- Only `/api/events` is covered by tests. The three `file-routes.ts` sites are fixed but untested — covering them needs session fixtures (`findSessionOrFail`), which felt out of proportion for a header spread. Reverting those three alone leaves the suite green.
- The inheritance logic now exists in three places (`sendRawStream`, the new helper, an inline copy in `server.ts`). The inline copy avoids the server importing from a route module it registers — but if you would rather have one shared util in a neutral module, say so and I will move it.
- Inheriting `Access-Control-Allow-Methods`/`-Headers`/`-Max-Age` onto non-preflight responses is harmless (browsers ignore them there); it is a consequence of inheriting wholesale rather than cherry-picking, and I judged the simpler rule to be the safer one.

---

### DIFF
--- a/.changeset/sse-inherits-security-headers.md
+++ b/.changeset/sse-inherits-security-headers.md
@@ -1,0 +1,16 @@
+---
+'aicodeman': patch
+---
+
+Routes that answer with `reply.raw.writeHead()` no longer drop the headers the
+security hook set.
+
+`writeHead` writes straight to the Node response and bypasses Fastify's header
+store, so everything the `onRequest` hook granted was silently lost — including the
+`Access-Control-Allow-Origin` it emits for localhost origins, and the
+`X-Content-Type-Options` / `X-Frame-Options` / CSP headers. A localhost page could
+therefore call every other `/api` endpoint cross-origin while its EventSource
+failed CORS.
+
+Affects `GET /api/events` and the three raw-writing routes in `file-routes.ts`
+(`file-raw`, `tail-file`, `download`).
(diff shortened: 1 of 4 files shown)


Ark0N:
Thanks Christian, this is a great catch. The writeHead bypass explanation is exactly right, and I appreciate that the test drives a real WebServer instead of an inline copy of the hook (and that you verified it fails with the fix reverted). Verified locally on top of the review: tsc clean, all 5 tests pass, CI green. Merging!

PR #225, opened 2026-08-07
fix(mux): bound the process-tree walk — unbounded pgrep recursion can take a machine down

christianhaberl:
## What happened

`getChildPids` ran `pgrep -P <pid>` per node and recursed with no visited set, no depth limit and no node cap. Two further sites forked a `pgrep` per session on every stats tick.

Across ~28 adopted tmux trees the fan-out exploded — and because each `pgrep` blocks in the kernel while reading `/proc/<pid>/cgroup` under WSL, none returned while the walk kept spawning more:

```
~13,000  pgrep processes stuck in D-state
~39,000  processes total
 >13,000 load average
```

The machine was only recoverable by restarting WSL, which cost every running session. Diagnostic commands timed out too — they read `/proc` as well.

This was on WSL, where the procfs read blocks hardest. The unbounded recursion itself is platform-independent; WSL is what turns it from slow into fatal.

## The change

- **One** `ps -eo pid=,ppid=` snapshot, briefly cached, refreshed **asynchronously** behind a single-flight guard. Async matters: under the same procfs pathology `execSync`'s timeout cannot return (`spawnSync` waits for the unkillable child), which would freeze the server where a hung async poll only costs staleness.
- The traversal moved to `proc-tree.ts` as a **pure function** — breadth-first, with a visited set (a stale snapshot can contain a cycle), a depth cap and a node cap, both reporting when they truncate. Pure so the tests exercise the shipped code rather than a copy.
- The kill path forces a fresh snapshot: the wait between SIGTERM and the survivor re-scan (200 ms) sits inside the cache TTL (2000 ms), so reading the cache there would return pre-SIGTERM state and aim SIGKILL at stale PIDs. That wait is bounded (1500 ms) so a wedged `ps` cannot stop `killSession` reaching its process-group and tmux fallbacks.
- Any `ps` error keeps the previous snapshot instead of caching partial output as fresh — a truncated table would make whole subtrees invisible, including to the kill path.

## Tests

13, including three that drive `TmuxManager` itself. **With the caps bypassed at the call site, 3 of them fail.**

The snapshot refresh is stubbed there, because otherwise the manager runs a real `ps`, replaces the fixture, and the test silently measures the machine's own process tree instead of the tree under test. An earlier version of this test passed with both caps bypassed for exactly that reason.

## Two known nits

- The depth-truncation warning fires once too often at the exact boundary: a tree whose deepest leaf sits at generation 10 returns a complete result but still logs `hit the 10-depth cap`. Cosmetic — the check tests `frontier.length > 0` rather than whether unseen children remain. Happy to tighten it if you would rather not have a false warning in the log.
- The 1500 ms rejection timer inside the `Promise.race` is not cleared when the refresh wins. It is a bounded, short-lived timer, but it does keep the event loop alive for up to 1.5 s past a fast path.

---

### DIFF
--- a/.changeset/bounded-process-tree-walk.md
+++ b/.changeset/bounded-process-tree-walk.md
@@ -1,0 +1,17 @@
+---
+'aicodeman': patch
+---
+
+Bound the process-tree walk that could take a machine down.
+
+`getChildPids` ran `pgrep -P <pid>` per node and recursed with no visited set, no
+depth limit and no node cap. Across ~28 adopted tmux trees the fan-out exploded,
+and because each `pgrep` blocks in the kernel while reading `/proc/<pid>/cgroup`
+under WSL, none returned while the walk kept spawning more — ~13,000 `pgrep`
+processes stuck in D-state out of ~39,000 total, load average above 13,000,
+recoverable only by restarting WSL.
+
+Now: one `ps` snapshot, breadth-first with a visited set, a depth cap and a node
+cap, in a pure module (`proc-tree.ts`) that the regression tests exercise
+directly. The snapshot is refreshed asynchronously, and the kill path forces a
+fresh one so the SIGKILL escalation cannot re-read pre-SIGTERM state.
--- a/src/proc-tree.ts
+++ b/src/proc-tree.ts
@@ -1,0 +1,87 @@
+/**
+ * @fileoverview Bounded descendant walk over a process-tree snapshot.
+ *
+ * Split out of `tmux-manager.ts` so the traversal can be unit-tested directly. It
+ * previously lived as a private method, which meant the regression test had to keep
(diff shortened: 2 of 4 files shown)


Ark0N:
Thanks Christian! This one really matters: the WSL fork-storm writeup is sobering, and the fix has exactly the right shape (one ps snapshot, pure bounded BFS that the tests exercise directly, async refresh so a wedged ps costs staleness instead of freezing the server, and the kill path forcing a fresh snapshot so SIGKILL never aims at pre-SIGTERM state). The two nits you flagged are cosmetic and fine as-is. Verified locally: tsc clean, the new proc-walk suite plus the existing tmux-manager suite pass (61 tests), CI green. Merging!

PR #226, opened 2026-08-07
fix(api,ws): an input whose delivery fails can be retried instead of being lost

christianhaberl:
## The bug

Both input paths recorded the `(clientId, seq)` pair as applied and acknowledged the frame **before** knowing whether the write had landed:

- `POST /api/sessions/:id/input`, because its mux write is fire-and-forget so the response never waits on a tmux child
- the WebSocket handler, because it ACKed unconditionally

When the write then failed, the client dropped the frame from its durable queue — correctly, per the protocol in `docs/architecture-invariants.md`: frames persist until the WS `{t:'ia',seq}` or an HTTP 2xx. The server then rejected the retry as a duplicate. **The reliable-delivery layer itself made the loss permanent.**

`Session.write()` returned `void`, so a session whose PTY was gone discarded the data with no signal at all.

## How it showed up

A prompt displayed as sent in a chat client, a 200 in the proxy log, and an empty prompt line in the pane.

## The change

- `forgetInputSeq()` rolls the bookkeeping back on failure — but only when that seq is still the newest; a later input has superseded it and must not be re-opened.
- The WebSocket handler withholds its ACK when the write did not land, so the client redelivers. A deduplicated frame is still ACKed, since the client must be able to drop what it already delivered.
- `Session.write()` reports whether it reached a PTY.

Response codes are **unchanged**, deliberately: a session can legitimately have no PTY yet (created, not started), and turning that into a failure status would be a contract change of its own. An earlier attempt at that broke `test/session.test.ts`, which writes to exactly such a session.

## What this does not fix

The root cause stays: the POST still answers 200 before the mux write is attempted, so a client that treats any 2xx as final cannot learn about that failure. Closing it would mean awaiting the tmux child inside the request — a separate decision, and yours to make.

What closes is the narrower window — the write failed **and** the ACK never reached the client — plus the whole WebSocket path.

Two consequences worth stating rather than letting you find:

- A withheld WS ACK only triggers redelivery on the client's **next** flush, so recovery is not immediate.
- On the POST mux path, a redelivery arriving while the first attempt is still in flight is ACKed as a duplicate; if the original then fails, the rollback re-opens the seq but the client has already dropped the frame. That sits inside the acknowledged root-cause window above and is not made worse by this change.

## Tests

12 across the two touched files. Verified in both directions, each half separately:

| Reverted | Failing |
|---|---|
| `src/web/routes/session-routes.ts` | 2 |
| `src/web/routes/ws-routes.ts` | 1 |

They drive the real HTTP route and a real listening WebSocket server, not the `Session` primitives alone — an earlier version tested only the primitives and stayed green with the fix removed.

---

### DIFF
--- a/.changeset/input-delivery-retryable.md
+++ b/.changeset/input-delivery-retryable.md
@@ -14,8 +14,11 @@ been delivered.
 
 The bookkeeping is now rolled back on failure and the WebSocket ACK withheld, so
 the client redelivers. `Session.write()` reports whether it reached a PTY at all
-instead of silently swallowing the data, and the non-mux POST branch — whose
-response has not gone out yet — answers `OPERATION_FAILED` rather than a cheerful 200.
+instead of silently swallowing the data.
+
+Response codes are unchanged: a session can legitimately have no PTY yet (created
+but not started), so turning that into a failure status would be a contract change
+of its own.
 
 Note this does not remove the root cause: the POST still answers 200 before the
 mux write is attempted, so a client that treats any 2xx as final still cannot
--- a/.changeset/input-delivery-retryable.md
+++ b/.changeset/input-delivery-retryable.md
@@ -1,0 +1,23 @@
+---
+'aicodeman': patch
+---
+
+An input whose delivery fails can be retried instead of being lost for good.
+
+Both input paths recorded the `(clientId, seq)` pair as applied and acknowledged
+the frame _before_ knowing whether the write had landed — the POST route because
+its mux write is fire-and-forget, the WebSocket handler because it ACKed
+unconditionally. When the write then failed, the client dropped the frame from its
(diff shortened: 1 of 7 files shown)


Ark0N:
Thanks Christian, another excellent one. Rolling back the dedup bookkeeping only when the seq is still newest is exactly the right guard, and withholding the WS ACK so the client's durable queue redelivers closes the loss window properly. Also appreciated: keeping the response codes unchanged (a session without a PTY yet is legitimate) and stating the remaining root-cause window honestly instead of papering over it. Verified locally: tsc clean, the new delivery tests plus ws-routes and session suites pass (50 tests), CI green. Merging!

Three solid, well-tested, well-argued fixes in a row: contributions like these are a joy to review. And I hear a tabs-on-the-side toggle might be coming next; very much looking forward to that one!

PR #244, opened 2026-08-09
fix(mobile): route terminal taps without breaking keyboard focus

Lint111:
Reopening #186 on its own branch, rebased on current master, as you suggested in #173.

This is the one you said you wanted most, and it now carries the fix for the blocker you reduced — the keyboard being unreachable — rather than just the tap routing.

## The blocker, fixed

You reported that tapping the terminal on a phone no longer raises the keyboard, so there is no way to type. That was real, and it was still present on my rebased branch until the last commit here.

Measured by dispatching real `touchstart`/`touchend` events against a running server, iPhone-class viewport (390×844), claude-mode session, tapping the middle of the terminal:

| | `document.activeElement` after the tap |
| --- | --- |
| master `b1614e8` | `textarea.xterm-helper-textarea` |
| this branch, before the fix commit | `body` |
| this branch, now | `textarea.xterm-helper-textarea` |

**Mechanism**, matching your reduction exactly: `_classifyMobileTerminalTap()` returns `content` for any non-prompt row, so `_handleMobileTerminalTap()` called `_blurMobileTerminalInput()`, while `touchstart`'s `preventDefault()` had already cancelled the compatibility click that would otherwise focus xterm. Both routes to focus closed on the same gesture.

## Why the blur could not simply be deleted

Two existing tests in `test/mobile/keyboard.test.ts` assert the blur, and they encode correct behaviour: tapping a collapsible readback or a `• Working (… esc to interrupt)` status row acts on the CLI, and popping the keyboard there is wrong. Your repro taps an inert transcript row, where the only sensible outcome is "let me type". Both classified as `content` and were treated identically — that conflation was the actual defect.

The fix distinguishes them by the **affordance the CLI prints** (`ctrl+r to expand`, `tap to collapse`, `esc to interrupt`) rather than by row titles. Titles vary per CLI and per version; the affordance is what makes a row actionable in the first place. Since the hint sits on its own row and a finger lands on the title, the adjacent row is consulted too.

I mention this because my first attempt keyed on title strings (`Agent readback`, `Tool result`) taken from the test fixtures. The full suite passed, and a realistically-worded codex readback (`Read src/foo.ts (120 lines)` / `ctrl+r to expand`) still swallowed the keyboard. It matched the fixtures, not the behaviour. Caught by probing a readback whose wording differs from the tests.

## On the test you asked for

You asked for a test asserting a tap ends with `document.activeElement === terminal.textarea`. It is here, dispatching a real gesture via `page.touchscreen.tap`.

**In fairness, it is not red on master.** Master does not have this bug, so the assertion is green there. It is a regression guard for this branch, not evidence against master. I would rather say that plainly than let the table above imply a comparison it does not support.

The same caveat applies to the unit tests in `test/terminal-touch-tap.test.ts`: 8 of 27 fail on master, but with `TypeError: ... is not a function`, because the methods are new. That only proves absence, not wrong output.

One methodological note that may be worth having in the repo: calling `_handleMobileTerminalTap()` directly reports the correct intent and looks fine, because it bypasses the `touchstart` `preventDefault()`. Only a dispatched touch gesture reproduces the bug. Any focus test here has to go through real events.

## Verification

Against master `b1614e8`:

```
npx tsc --noEmit                                  clean
test/terminal-touch-tap.test.ts                   27 passed (27)
npm run test:mobile -- test/mobile/keyboard.test.ts   5 failed | 35 passed (40)
```

The 5 failures are pre-existing and identical on master (stale layout/accessory-bar expectations and a CJK timeout); they are untouched here. Full mobile suite, this branch vs pristine master on the same machine: **both 54 failed**, identical per-file counts, branch running 356 tests to master's 352 — the added tests pass and nothing regresses.

## Coexistence with #205

The touch-forwarding path added in #205 is unchanged. Drag gestures still route through `_shouldForwardTouchScrollToApp()` — forwarded as SGR wheel events for verified Claude versions, local xterm scrollback otherwise. This PR only changes what happens on a **tap** (no scroll), and only which rows dismiss the keyboard. The SGR mouse report a content tap sends is byte-identical to before; the readback test still asserts exactly one report of the form `\x1b[<0;N;1M\x1b[<0;N;1m`.

## Scope

3 files. `src/web/public/terminal-ui.js` plus the two test files. No changes to any guard or policy test.

### DIFF
--- a/src/web/public/terminal-ui.js
+++ b/src/web/public/terminal-ui.js
@@ -653,6 +653,7 @@ Object.assign(CodemanApp.prototype, {
       let didScroll = false; // track whether touchmove fired (tap vs scroll)
       let touchStartY = 0;
       let tapStartedWithTerminalFocus = false;
+      let tapStartIntentCache = null;
       const TAP_THRESHOLD = 8; // px — ignore micro-drift to distinguish tap from scroll
       container.addEventListener(
         'touchstart',
@@ -666,11 +667,24 @@ Object.assign(CodemanApp.prototype, {
             isTouching = true;
             didScroll = false;
             tapStartedWithTerminalFocus = this._isMobileTerminalInputFocused();
+            // Classifying scans the whole viewport with translateToString, and
+            // this runs at the start of EVERY gesture including scroll drags.
+            // Cache the result for the touchend of this same gesture rather than
+            // recomputing it; the cache is keyed on the exact start coordinates
+            // so a finger that moved re-classifies at its real position.
             const touchStartIntent = this._classifyMobileTerminalTap(touchLastX, touchLastY);
-            if (touchStartIntent !== 'input') {
+            tapStartIntentCache = { x: touchLastX, y: touchLastY, intent: touchStartIntent };
+            if (touchStartIntent === 'content') {
               // Cancel xterm/browser focus before the compatibility click can
               // open the OS keyboard. Content taps are re-emitted as SGR on
-              // touchend; history taps deliberately remain inert.
+              // touchend.
+              //
+              // 'history' is deliberately NOT included. A scrolled-up viewport
+              // sends nothing, so there is no compatibility click worth
+              // cancelling — and preventDefault() here, paired with touchend's
+              // early return, closes both routes to focus at once. Since
+              // selectSession() ends with scrollToLastNonEmptyLine(), that made
+              // the keyboard unreachable after every tab switch.
               ev.preventDefault();
               this._blurMobileTerminalInput();
             }
@@ -700,7 +714,6 @@ Object.assign(CodemanApp.prototype, {
             // fling, so a jittery tap would both position the cursor AND scroll.
             if (!didScroll) return;
             ev.preventDefault();
-            touchLastX = ev.touches[0].clientX;
             const delta = touchLastY - touchY; // positive = scroll down
             pixelAccum += delta;
             velocity = delta * 1.2;
@@ -737,7 +750,13 @@ Object.assign(CodemanApp.prototype, {
             const touch = ev.changedTouches && ev.changedTouches[0];
             if (touch) {
               this._suppressTrustedTapMouseEvents();
-              this._handleMobileTerminalTap(touch, tapStartedWithTerminalFocus);
+              const cached =
+                tapStartIntentCache &&
+                tapStartIntentCache.x === touch.clientX &&
+                tapStartIntentCache.y === touch.clientY
+                  ? tapStartIntentCache.intent
+                  : null;
+              this._handleMobileTerminalTap(touch, tapStartedWithTerminalFocus, cached);
             }
           }
           tapStartedWithTerminalFocus = false;
@@ -3371,9 +3390,6 @@ Object.assign(CodemanApp.prototype, {
     const mouseTrackingOn = !!mouseMode && mouseMode !== 'none';
     if (!mouseTrackingOn && !this._sessionUsesServerMouseStrip()) return 'input';
 
-    // Permission/elicitation prompts own the full live terminal until answered.
-    if (document.body?.classList?.contains('terminal-action-pending')) return 'content';
-
     const buffer = this.terminal.buffer?.active;
     if (!buffer?.getLine) return 'input';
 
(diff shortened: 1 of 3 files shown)


Ark0N (changes requested):
Thanks for the rebase and for the write-up, particularly the note that the focus test is not red on master and the methodological point that only a dispatched touch gesture reproduces the bug. Both are correct and worth keeping in the repo.

I fetched the branch into an isolated worktree, ran the checks, and probed the behaviour with real touch gestures against a live server. Requesting changes: the blocker is still reachable, through a second path this PR does not cover.

## 1. Blocking: the keyboard is still unreachable after a tab switch

`selectSession()` ends with `scrollToLastNonEmptyLine()` (`app.js:4670`, also `:4442` on snapshot restore). That parks the viewport above the bottom whenever the buffer is taller than the screen and ends in blank rows, which is every real session. `_classifyMobileTerminalTap()` opens with:

```js
if (!this._terminalViewportAtBottom()) return 'history';
```

so every tap classifies as `history`, touchstart runs `preventDefault()` + `_blurMobileTerminalInput()`, and touchend's early return skips `_focusMobileTerminalInput()`. Both routes to focus close on the same gesture: the exact mechanism the last commit fixed for inert rows.

Measured, iPhone-class viewport (390x844), claude-mode session, buffer taller than the screen, `scrollToLastNonEmptyLine()` then a real `page.touchscreen.tap`:

| | viewport state | `document.activeElement` after the tap |
| --- | --- | --- |
| master `b1614e8` | `viewportY 82 / baseY 83` (off-bottom) | `textarea.xterm-helper-textarea` |
| this branch | `viewportY 82 / baseY 83` (off-bottom) | `body` |

Same result when the tap lands on the prompt row itself. Master is unaffected because its touchend calls `this.terminal.focus()` unconditionally; only `_sendSyntheticSgrTap` was gated on at-bottom.

Suppressing the mouse report while scrolled up is right, and an improvement over master. Blurring is not, and there is no reason to `preventDefault()` at touchstart for a gesture that will send nothing.

Repro, as a `test/mobile` case:

```ts
// setup: claude-mode session, write ~120 transcript rows + a prompt box + 2 trailing blanks
await page.evaluate(() => {
  app.scrollToLastNonEmptyLine();          // what selectSession() does on every tab switch
  (document.activeElement as HTMLElement | null)?.blur?.();
});
await page.touchscreen.tap(x, y);          // anywhere in the terminal, prompt row included
const activeClass = await page.evaluate(() => document.activeElement?.className);
expect(activeClass).toContain('xterm-helper-textarea');   // green on master, red here
```

## 2. `terminal-action-pending` does not exist

Both new functions guard on it:

```js
if (document.body?.classList?.contains('terminal-action-pending')) return 'content'; // :3117
if (document.body?.classList?.contains('terminal-action-pending')) return true;      // :3216
```

`grep -rn terminal-action-pending src test docs` returns nothing on master or on this branch. Both branches are permanently false, so "Permission/elicitation prompts own the full live terminal until answered" never engages. Either wire the class up where prompts are detected, or drop the guards so the comment does not promise coverage that is not there.

## 3. The claude status literal does not match claude

`/^\s*[•·]\s*Working\b/` in `_isActionableMobileTerminalTap`, and `/^\s*[•·]\s*Working\b.*(?:background|esc to interrupt)/i` in `_classifyMobileTerminalTap`, are fixture-shaped. Live panes here on claude 2.1.226 print:

```
✻ Cooked for 2m 6s
✻ Baked for 9m 47s
✻ Brewed for 18m 41s
```

Different bullet (`✻`, U+273B, not `•`/`·`) and a randomised verb that is never "Working". This is the same trap your PR body documents about title strings, one layer down. The affordance regex does catch the actively-working row via "esc to interrupt", so behaviour largely survives, but both `Working` clauses are dead code.

## 4. `_shouldForwardTouchScrollToApp()` is never called

Defined at `:3488` and unit-tested, but unreachable: touchmove still calls `_shouldForwardWheelToApp({ shiftKey: false })` directly, exactly as on master. "Drag gestures still route through `_shouldForwardTouchScrollToApp()`" in the PR body is not accurate, and wiring it as written would restrict forwarding to claude only, dropping gemini from the path #205 established. Please either wire it deliberately (with that mode change justified) or drop it and its test.

## Smaller items

- **Undisclosed behaviour change.** `shouldActivate = intent === 'content' || startedWithTerminalFocus` means the first tap on the prompt row no longer positions the cursor; it takes two taps now. That looks deliberate (the "keeps the first prompt tap focus-only" test pins it), but it should be in the PR description, since it changes an interaction people already rely on.
- **Cost per gesture.** `_classifyMobileTerminalTap` does a full-viewport `translateToString` scan. It now runs at touchstart (which also became non-passive), again inside `_handleMobileTerminalTap`, and `_isActionableMobileTerminalTap` does a third. Every scroll drag, not just every tap, pays a screen scan at gesture start. The touchstart result is already computed and then discarded; cache it for the touchend pair.
- **Duplicate line.** `touchLastX = ev.touches[0].clientX;` now appears twice in the touchmove handler.
- **Affordance false positives.** `/\b(?:ctrl\+\w+|tap|click|enter|esc)\b[^.]{0,24}\bto\s+(?:expand|collapse|view|open|interrupt|see)\b/i` matches ordinary transcript prose such as "click here to open the file", which would dismiss the keyboard on an inert row.
- `_handleMobileTerminalTap` returns `'history'` for the "no touch or no terminal" bail-out, conflating a real classification with a guard.

## What checks out

Everything you claimed verifies:

```
npx tsc --noEmit                              clean
npm run lint                                  clean
npm run format:check                          clean
npm run check:frontend-syntax                 28 files parse cleanly
test/terminal-touch-tap.test.ts               27 passed (27)
test/mobile/keyboard.test.ts   branch         5 failed | 35 passed (40)
test/mobile/keyboard.test.ts   master b1614e8 5 failed | 30 passed (35)
```

Identical failing test names on both sides, so the 5 are pre-existing and untouched here, as you said.

The direction is right, and separating "TUI owns this row" from "the user wants to type" is the correct framing. Item 1 is the one that has to be fixed before this can land; 2 through 4 are code that does not do what it says.

PR #251, opened 2026-08-09
feat(cases): clone a Git repository as a new case (#236)

### LINKED ISSUE #236: feat: clone a public Git repository as a new case
## Motivation

The Add Case modal currently supports creating a blank case and linking an existing directory. It would be useful to create a case directly from a public Git repository without first cloning it manually.

## Proposed experience

Add a **Clone Repository** tab to the Add Case modal with:

- Repository URL
- Case name
- Optional branch or tag

Submitting the form would clone the repository into the current user's normal `codeman-cases/<name>` directory, register it as a local case, refresh the case list, and select the newly created case.

For the initial version, cloning would be deliberately synchronous from the application's perspective: the HTTP request stays open while `git clone` runs, the submit button shows a "Cloning..." state, and the response returns only when cloning succeeds, fails, or times out. This proposal does **not** introduce a background-job system, progress streaming, polling, or cancellation machinery.

## Initial scope

- Support public repositories that can be cloned non-interactively
- Use the host's installed `git`
- Do not accept, store, or manage credentials or access tokens
- Do not automatically start an agent session
- Treat the cloned directory as a created case, so deleting the case deletes its working tree after the existing confirmation
- Allow an optional branch/tag, mapped to `git clone --branch <ref> --single-branch`

"Public repositories only" here means Codeman provides no authentication facility. If the host's existing Git configuration happens to supply credentials, Codeman would neither collect nor manage them.

## Possible implementation shape

- Add `POST /api/cases/clone`
- Add a dedicated Zod request schema
- Resolve and validate the destination using the existing per-user cases-directory logic
- Reject duplicate case names before invoking Git
- Invoke Git without a shell, using an argument array and `--` before repository/destination operands
- Reject dangerous Git transports such as `ext::`
- Set `GIT_TERMINAL_PROMPT=0` so the request cannot hang on an invisible prompt
- Apply a bounded timeout and capture bounded stderr for a useful error response
- If cloning fails, remove only the exact destination created for that attempt
- Preserve repository contents rather than overwriting an existing `CLAUDE.md`
- Broadcast the normal case-created event and refresh/select the case in the UI
- Add focused helper, route, and UI tests

Possible request body:

```json
{
  "name": "my-project",
  "repository": "https://github.com/example/project.git",
  "ref": "main"
}
```

## Out of scope for the initial version

- Private-repository authentication
- Entering or storing Git credentials in Codeman
- Background jobs
- Clone progress streaming or polling
- Cancellation UI
- Submodule initialization
- Automatically launching a session
- Advanced clone options such as depth or sparse checkout

## Questions

1. Does cloning into the managed cases directory fit the intended case model?
2. Should this be a separate tab, or an option within **Create New**?
3. Should cloned cases have the same deletion semantics as other created cases?
4. Should Codeman add its hook configuration immediately after cloning, or let the existing session-start scaffolding handle that?
5. Is an optional branch/tag useful for the first version, or should the initial UI accept only a repository URL and case name?
6. Is holding the request open until the bounded clone completes acceptable for the first version?

If this direction fits the project, I'd be interested in implementing it. Feedback on the scope and UX would be very welcome before I begin.

Ark0N:
> ⏸ **Not merging yet.** #236 is @DodgyBadger's proposal, so this is held open for their review first. See the comment below.

Implements #236 (proposed by @DodgyBadger): an Add Case → **Clone Repo** tab that clones a public repository straight into `codeman-cases/<name>` and registers it as a normal local case.

Closes #236.

## What it adds

**`POST /api/cases/clone`** — synchronous by design, exactly as proposed: the request stays open for the clone (bounded by `GIT_CLONE_TIMEOUT_MS`, default 5 min), so there is no job store, no polling and no cancellation surface. Success broadcasts the usual `case:created`, so the case still appears if a proxy idle-timeout kills the request while git is still working.

**`POST /api/cases/clone-preflight`** — runs `git ls-remote --symref` so the UI can answer, while you are still typing, whether the URL is cloneable *without credentials*, what its default branch is, and which branches/tags exist.

**Add Case → Clone Repo tab** — live verdict under the URL field, case name filled from the parsed repo (a name you type wins), branch/tag as a datalist of the remote's real refs, optional shallow clone, and a **Brain** picker (installed CLIs only) that points the Run button at the chosen agent when the clone finishes. Starting a session stays opt-in. The tab hides itself when the server has no `git`.

## Security notes for review

Core is `src/git-clone.ts`, split into a pure half (URL parse, argv/env, `ls-remote` parse, stderr classification) and a thin IO half, so every decision below is unit-testable without spawning anything.

- **`<name>::<payload>` transports are refused as a family, not by name.** `ext::sh -c <cmd>` is the famous one, but any of them dispatches to a `git-remote-<name>` helper, which turns a clone into arbitrary command execution.
- **A leading `-` is refused AND `--` precedes every operand.** Either alone is one edit away from being a hole.
- **argv arrays, never a shell**, so quoting is not part of the threat model here. URLs carrying `user:password@` are refused outright.
- **Non-interactive is more than `GIT_TERMINAL_PROMPT=0`.** `gitNonInteractiveEnv()` also empties `GIT_ASKPASS`/`SSH_ASKPASS`, sets `SSH_ASKPASS_REQUIRE=never` and an empty `DISPLAY`, `GCM_INTERACTIVE=never`, and `ssh -oBatchMode=yes`. With the request held open, any one of those left open is a hang instead of an error. `HOME`/`PATH` stay inherited on purpose, so a user's own credential helper or ssh agent keeps working; Codeman itself collects and stores nothing.
- **The timeout signals the process GROUP.** `git clone` fans out into `git-remote-https`/`index-pack`; SIGTERM to the parent alone can leave the fetch running, so it spawns detached and escalates SIGTERM → SIGKILL to `-pid`.
- **Bounded output and concurrency**: credential-redacted, control-stripped stderr *tail*; capped `ls-remote` stdout; 500 refs of each kind; and a global 2-op pool (`CODEMAN_MAX_GIT_OPERATIONS`), since N simultaneous large clones is a resource-exhaustion shape.
- **Repository contents beat scaffolding**: an existing `CLAUDE.md` is kept, hooks are merged into whatever `.claude/settings.local.json` the repo shipped, and a repo shipping its own `.claude/settings*` is reported back as a warning, because those hooks run locally as soon as a session starts in the case.
- **Failure leaves nothing behind**: only the directory this attempt created is removed, and a pre-existing destination is refused rather than cloned into, so a failed clone can never squat on a case name.
- **Multi-user**: not admin-gated, unlike `/api/cases/link`, because it writes only inside the caller's own `resolveCasesDir`. The exception is a local-path/`file://` source, admin-only there, since per-user spaces live inside one `$HOME` and a local clone would read straight through that boundary.

Two behaviours worth knowing rather than rediscovering: hosts answer `Repository not found` for a private repo *and* for a typo when unauthenticated, so the message says both; and git's *first* stderr line is `Cloning into '<dest>'...`, so the error detail takes the last `fatal:`/`error:`/`remote:` line instead (quoting the first reported the destination path when the branch was wrong).

## Verification

- `test/git-clone.test.ts` — the pure half exhaustively (every refusal has its own case) plus **real `git` against a real local bare repo** for clone / single-branch / tag+shallow / bad-ref cleanup / pre-existing destination / timeout-kill / slot-release.
- `test/routes/case-clone-routes.test.ts` — **unmocked fs**, clones through the endpoint for real; asserts the repo's own `CLAUDE.md` survives, hooks land beside repo settings, and a failed clone leaves no case directory.
- `test:ci` green (one unrelated `EADDRINUSE` from a concurrent local run; that suite passes alone). Typecheck, lint, prettier, frontend-syntax and public-asset checks all clean.
- Manually against a live instance: real clones of `octocat/Hello-World` (plain, at a tag, shallow), duplicate name → `ALREADY_EXISTS`, bad ref → 400 with no leftover directory, `ext::` refused with nothing executed; Playwright on desktop + iPhone 13 for the verdict/autofill/datalist, tab wrapping, and the full submit path including the optional session start.

No changeset here on purpose, so `Release` stays green until the next COM.


DodgyBadger (changes requested):
Strong implementation overall. A few hardening issues need addressing before merge:

- Post-clone scaffolding follows repository-controlled symlinks (`CLAUDE.md`, `.claude`, `settings.local.json`), allowing writes outside the cloned case. Reject/skip symlinked scaffold paths and verify confinement.
- Failed-clone cleanup can race: two requests may both see a free destination, then the loser unconditionally removes the winner’s tree. Clone into an attempt-owned temp dir and atomically rename; only clean up that temp dir.
- `decodeURIComponent(url.pathname)` can throw on malformed escapes (for example `%`), producing a 500 instead of `BAD_SYNTAX`.
- The Git concurrency pool has an unbounded waiter queue, and operation timeouts begin only after acquisition. Bound/reject queued work and include queue time in the deadline.

The URL/transport policy and preflight shape otherwise look good.

PR #376, opened 2026-09-03
feat: auto-name sessions from first prompt

shenlvkang-collab:
## Problem
New sessions are initially named with generic project/sequence labels, so it is difficult to identify their purpose in the session manager.

## Solution
- Derive a bounded local title from the first submitted prompt.
- Ignore slash commands and terminal control sequences.
- Persist whether a name is auto or manual.
- Never overwrite an explicit rename.
- Update mux metadata and broadcast the new name immediately.

No prompt text is sent to an additional model or written to logs.

## Verification
- typecheck
- lint
- build
- frontend syntax check
- focused input/listener tests: 23/23 passed

The implementation was also adapted and validated against the older Codeman deployment on silverdou.

### DIFF
--- a/docs/wiki/The-Dashboard.md
+++ b/docs/wiki/The-Dashboard.md
@@ -66,6 +66,14 @@ reloading while a permission prompt is blocking does not lose the red tab.
 
 Tabs can also be dragged to reorder.
 
+### Automatic session names
+
+New sessions start with a short project/sequence name so they can be created immediately.
+After the first task prompt is submitted, Codeman replaces that placeholder with a short
+title derived locally from the prompt's first sentence. Slash commands such as `/clear` do
+not become titles. A name you set with the inline rename action is treated as manual and is
+never overwritten by automatic naming.
+
 On phones the strip scrolls horizontally instead of wrapping, and the active tab is always
 scrolled into view. It is not reordered to the front, so the `Alt+N` numbering stays stable.
 
--- a/src/session-auto-name.ts
+++ b/src/session-auto-name.ts
@@ -1,0 +1,101 @@
+/**
+ * Helpers for assigning a useful default name after the first submitted prompt.
+ *
+ * This deliberately does not call an LLM: the prompt is already available at
+ * the input boundary, so a bounded local title is private, deterministic, and
+ * works for every CLI backend.
+ */
+
+const MAX_PROMPT_BUFFER_LENGTH = 8_192;
+const MAX_AUTO_NAME_CODE_POINTS = 72;
+
+/**
+ * Tracks terminal input until Enter is received. Terminal input arrives in
+ * arbitrary chunks, so this keeps only a small composer buffer and ignores
+ * navigation/control escape sequences.
+ */
+export class SubmittedPromptTracker {
+  private buffer = '';
+  private escapeSequence = '';
+
+  feed(data: string): string[] {
+    const submitted: string[] = [];
+
+    for (const character of data) {
+      if (this.escapeSequence) {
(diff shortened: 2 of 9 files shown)


Ark0N:
Checking in on this one, since I am assembling the next release and this is not in it.

Nothing has changed here since my review on 7 September, and `format:check` is still red on `src/session.ts:686` (one `npx prettier --write src/session.ts` away). I had a second reviewer go over it independently in the meantime, with a different model, and it reached the same findings I did without seeing my comment first: renaming on every prompt rather than the first, and the tracker sitting on the raw byte stream so Codeman's own programmatic writes count. The auto-resume nudge renaming a session to `continue` and an Approvals keystroke renaming it to `1` are both still live.

I am not asking you to redo the design. The ownership plumbing, the listener wiring and the restore path are fine and can stay exactly as they are.

What I do need before this can move is an answer to the question at the end of that review, because it decides how much of the rest you rework: a synced `autoNameSessions` setting, and whether you agree with the `w1-case: description` prefix form. `parseSessionPrefix()` already renders that as just the description with the prefix in the tooltip, and `_nextCaseSessionStartNumber()` still matches it, so the prefix keeps the case identity and the `w<n>` counter that a full name replacement breaks. If you think the bare title is better, say why and I will listen; I would rather hear the argument than have you build something you think is worse.

Happy to keep this open as long as you want to work on it. If you have moved on, that is completely fine too, just say so and I will close it rather than leave it sitting.

Ark0N:
Merged, thanks @shenlvkang-collab. The idea was worth landing, and the ownership plumbing you built (`nameSource`, the listener wiring, the restore path) is in as you wrote it, with one extra state.

Since the review had no reply I finished the rework myself in 5b920cb4 rather than leave the PR sitting:

- **First prompt only.** `applyAutoName()` now flips a placeholder to `auto` whether or not the string changed, so a later "1" cannot rename the tab. `nameSource` is `placeholder | auto | manual`; the rename route is the only manual path.
- **User input only.** `write()` / `writeViaMux()` take `{ fromUser: true }`, set by the browser WebSocket path and `POST /input` only. Ralph, respawn, cron, approvals and the trust-dialog keys never name a tab. Shells are excluded through the registry's `startMode` capability, and the send-key route feeds its Shift+Enter line feed to the tracker so a two-line prompt keeps its separator.
- **Prefix form.** The tab becomes `w3-case: fix the login redirect`, which the strip already renders as the title with the prefix in the tooltip, and the `w<n>` counter keeps working. That was the one question left open on the review; I went with the prefix because losing the case identity was the part I could not ship.
- **Tracker rules per key.** A bare Esc is resolved at the end of the chunk it arrives in; mouse reports, Tab, cursor keys and Shift+Tab are no-ops; Up/Down and Ctrl+P/N/R taint the draft so Enter submits nothing instead of a fragment; bracketed-paste newlines join with a space; the draft keeps its head past 8192 code points. Title heuristics: slash commands by shape (a path is a prompt), `!` escapes refused, first sentence only past 8 code points.
- **Opt-in.** Synced `autoNameSessions` setting, default off, in App Settings under Appearance, Tabs, because the prompt lands in `mux-sessions.json`, every `session:updated` and `/api/search`.

Tests live in `test/session-auto-name.test.ts`, the wiring test and a route test for the rename flip. Verified live on an isolated instance: prompts through the API and typed in a browser name the tab, a second prompt does not, shells and renamed tabs stay untouched, and `nameSource` survives a restart. Docs are in CLAUDE.md, the architecture invariants and the wiki.

It ships in the next release, with you in the Thanks.

PR #431, opened 2026-09-15
fix(terminal): four silent-failure paths — renderer freeze, replay race, reconnect gap, unbounded fetches

rounakdatta:
Four fixes for ways the mobile terminal can silently stop being correct — where "silently" is the operative word: in every case the buffer keeps updating, nothing throws, and the user's only recourse is a reload.

They came out of reading [Big-Pony/pocketshell](https://github.com/Big-Pony/pocketshell), a mobile-first terminal that has done unusually deep forensics on xterm's failure modes. Each item below was re-verified against Codeman's own source before being written — several things that repo warns about turned out to be already handled here, and those are not in this PR.

---

### 1. Renderer freeze after backgrounding (`terminal-ui.js`)

iOS **discards** scheduled `requestAnimationFrame` callbacks when a PWA goes to the background — not deferred, never delivered. xterm's `RenderDebouncer` only clears its `_animationFrame` handle from *inside* that callback:

```js
refresh() {
  if (this._animationFrame !== undefined) return;   // <- stale forever after
  this._animationFrame = requestAnimationFrame(() => this._innerRefresh());
}
_innerRefresh() { this._animationFrame = undefined; ... }   // never runs
```

One dropped callback leaves the handle permanently set, so every later render request returns on line one. Parsing is decoupled from rendering, so bytes keep filling the buffer correctly and nothing errors — the terminal is simply frozen. Closing and reopening fixes it because that constructs a new `Terminal`.

Codeman is **more** exposed than an app that mounts a terminal per session: there is exactly one xterm instance for the whole page load, so a single backgrounding can wedge it until a reload.

`_startRenderLivenessWatchdog()` polls every 2s and, when a *visible* terminal has been written to but has produced no frame for 4s, `_kickRenderer()` does what the dropped `_innerRefresh` would have: cancel the stale handle, clear the field, force a repaint.

⚠️ This reads xterm privates — there is no public API for any of it. Every access is optional-chained and wrapped, so a shape change upstream degrades to a no-op rather than throwing on a timer. `_renderService` only exists after `open()` (needs a real DOM), so the gate can't assert the field path.

`test/xterm-private-api.test.ts` now pins the **resolved lockfile version** rather than the declared `^6.0.0` range — the range was the wrong assertion in both directions, since a real upgrade to 6.4.0 could rename a private field while resolving inside it, and an innocuous range edit would fail while changing nothing installed.

Thank you for independently verifying the field path and the heal against 6.0.0 under jsdom — that covers the half of this I couldn't reach. The remaining unverified half is the *premise*: that iOS discards scheduled rAF callbacks on backgrounding, which is what leaves the handle stale. That still wants a real device.

### 2. Replay clears race live output (`app.js`)

xterm's `write()` is asynchronously queued; `Terminal.reset()` is synchronous and, per upstream's own docs, *"does not clear input buffers and does not reset the parser, thus the terminal will continue to apply pending input data."* Bytes queued just before a reset are parsed **after** it and fuse into the snapshot written next.

Reproduced against the real xterm 6 in this repo:

```
write('p8'); reset(); write('rmissions')   ->  "p8rmissions"
write('p8'); write('\x1bc'); write('rmissions')  ->  "rmissions"
```

`_resetTerminalForReplay()` already got this right by following the sync `reset()` with a *queued* `\x1b[3J\x1b[H\x1b[2J`. Two other paths — the `needsRefresh` reload and the `clearTerminal` refresh — hand-rolled `clear()` + `reset()` with no in-stream erase and were genuinely exposed.

All three now go through one function, which is a single queued `\x1bc` (RIS). RIS rather than the erase because `3J/H/2J` leaves modes, charsets, scroll regions and SGR state alone, so leftover bytes can park the terminal in alt-screen and survive the clear. Callers can still chunk the content — ordering in the queue is what matters, not writing it in one call.

### 3. Output lost when the WebSocket dies but SSE stays up (`app.js`)

The terminal WS protocol is asymmetric in a way that's easy to miss, because one half is excellent. **Input** frames carry `seq` + `cid`, the server applies each pair at-most-once and ACKs, and the client holds a durable queue until the ACK lands. **Output** frames (`{"t":"o","d":...}`) carry nothing, and `ws.onopen` re-sends dimensions and flushes queued *input* only.

I traced every emitter of `needsRefresh`, the one server signal that could cover the gap — there are two: external-CLI startup 3s after spawn (`session.ts`), and SSE backpressure drain (`sse-stream-manager.ts`). Neither fires on a WS reconnect.

**Scope correction from review:** an earlier draft of this described the gap as "output produced while a phone is off-network." That was wrong. If the network drops, SSE drops with it, and `handleInit`'s `keepTerminal` branch already calls `_onSessionNeedsRefresh` for exactly this reason. The genuinely uncovered case is narrower: the **WS dying while SSE stays up** — a half-open socket, a proxy idle-timeout, a ping timeout. `_onSSETerminal` discards every SSE terminal frame while `_wsReady` is true, and `_wsReady` only flips in `ws.onclose`, so detecting a half-open socket takes up to the ping+pong window and that whole span produces output nothing writes to the terminal.

Still worth closing, and this is the interim fix rather than the real one: reaching `onclose` at all means the drop was unintentional (`_disconnectWs` nulls the handler first), so the session is marked and the next successful open reconciles from the server's buffer. Sequencing the output frames properly is the follow-up.

`_markTerminalBufferReconciled()` clears the marker from every path that repaints that session's buffer — `_onSessionNeedsRefresh`'s `finally`, `selectSession` after its load, and `_cleanupSessionData`. Without that, `selectSession` (which loads the buffer and only *then* calls `_connectWs`) would have the socket replay the whole buffer a second time on top of the one just written.

### 4. Terminal captures had no deadline (`app.js`)

No fetch in the frontend carried a timeout — no `AbortController`, no `AbortSignal` — including `?full=1`, which `_maybeRefetchFullHistory` itself describes as *"unbounded-ish work: at the default history limit it can be megabytes."*

**Corrected from review:** the first version of this cleared its abort timer in a `finally` around `await fetch(url)` — which settles on response **headers**, not the body. So it only ever bounded a server that accepts a connection and never replies; the multi-megabyte body it was written for stayed unbounded. Reproduced against a server that sends headers immediately and stalls the body 4s under a 1s deadline: fetch resolved at 30ms, the timer cleared there, body completed at **4026ms unaborted**.

`_fetchTerminalCapture` now reads the body inside the helper and returns `{json, headers, headersAt}` — `headers` because two callers read `server-timing`, `headersAt` because those same callers measure header-vs-body time and can no longer observe that moment themselves. `_terminalCaptureInflight` is scoped the same way, so a body still streaming counts toward the budget of a capture starting beside it. Same test now aborts at 1005ms.

The budget scales with full-vs-tail and with captures already in flight, and degrades to a plain fetch where `AbortController` is missing — the deadline is a safety net, not a dependency.

### Also included

- **Service worker precache was dead, and my first fix made it worse.** The build content-hashes assets and rewrites `index.html`, but `sw.js` was maintained by hand with the pre-hash names, so every entry 404'd and `cache.add(...).catch(() => {})` hid it (**15 of 23**, verified against a running instance). Offline still worked, because the fetch handler caches every successful GET at runtime.

  **Corrected from review:** deriving the list from the manifest was not sufficient. `renderIndexHtml` runs `cacheBustAssets`, which appends `?v=<mtime>` to every same-origin `.js`/`.css` reference *including content-hashed names* — confirmed on a live instance: `vendor/xterm-zerolag-input.6fee72f2.js?v=1789402869101`. `caches.match` is query-sensitive, so a precache keyed on the bare path could never be hit, and my change turned entries that failed cheaply into ~1.3MB downloaded at every install that nothing could ever read back — once per deploy, since `CACHE_NAME` now rotates. The fallback match now passes `{ ignoreSearch: true }`, which also lets runtime-cached entries survive an mtime change. Pinned by a test, since it is subtle enough to be tidied away.

  The per-build `CACHE_NAME` stands on its own: `activate` deletes every cache that is not the current one, so the old constant `'codeman-v1'` meant that cleanup never ran and assets from every past release accumulated forever.
- **Crash-trail hygiene.** Entries are joined with `\n` into one localStorage value and beaconed, and at least one call site interpolates a server-controlled WS close `reason` — a newline there forges entries. Now flattened and length-capped.

---

### Testing

- Full gate green apart from `cron-service.test.ts` "blocks a sensitive system file via the blocklist", which fails identically on clean `origin/master` in my sandbox and — per your run — does not reproduce on yours. Environmental, not a regression.
- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run check:frontend-syntax`, `npm run check:public-assets` — all green.
- New tests: `test/terminal-resilience.test.ts`, `test/sw-precache-manifest.test.ts`, `test/xterm-private-api.test.ts`. All run in the gate, which is deliberate given the mobile suite is excluded from CI.
- The deadline fix has a **behavioural** test against a real socket (a server that stalls its body), plus a source guard asserting `await res.json()` runs before the `finally` — I verified that guard fails when the helper is reverted to the old shape, so it is not vacuous.
- `test/sw-precache-manifest.test.ts` now **parses** `HASHABLE` out of `scripts/build.mjs` instead of hand-copying it. You were right that the copy was the same drift this PR exists to fix; the parse is guarded against silently matching nothing.
- §2's RIS behaviour was verified against the real xterm build headless, not reasoned from docs.

### Not verified here

Still no browser on my machine, so no runtime verification from my side. With your jsdom check of the field path and the heal, what remains unverified is the **premise** of §1 — that iOS discards scheduled rAF callbacks on backgrounding — and the §3 reconnect path, which is traced through code rather than observed on a device. Both are worth the real-device pass you suggested before the release that carries this.

### On the split

Agreed, and noted for next time — the service worker and build script are a different subsystem from the four terminal fixes, and the only build-breaking risk in the set. If you would still prefer them separated before merging, say the word and I will split the sw/build commit onto its own branch and rebase this one to the terminal work alone.


Ark0N:
Thanks for this, and for the forensics write-up that goes with it. Four terminal silent-failure paths plus the service-worker precache is a lot of careful reading, and the replay-clear item in particular fixes a real corruption path nothing else here covered.

One thing worth saying up front: I verified the piece you flagged as most wanting eyes. Driving the repo's own xterm 6.0.0 under jsdom, `_core._renderService._renderDebouncer` exists after `open()`, `refreshRows` is a function, a forced stale `_animationFrame` genuinely makes `refreshRows` a no-op, and your kick sequence schedules a fresh frame. The field path and the heal are correct for the pinned version. I also reproduced the reset() race: `write('p8'); reset(); write('rmissions')` gives `p8rmissions`, the queued `\x1bc` gives `rmissions`, and RIS clears scrollback and leaves the alt buffer as you describe.

Two things need a change before this lands.

**1. The fetch deadline only covers time-to-headers (src/web/public/app.js:2543).** `_fetchTerminalCapture` clears its timer in a `finally` that runs when `await fetch(url)` settles, and a fetch promise settles on response headers. Every caller then reads the body with `await res.json()` with the deadline already gone. I checked it against a local server that sends headers immediately and stalls the body for 4s with a 1s deadline: fetch resolved at 20ms, the timer was cleared there, the body completed at 4018ms unaborted. So the multi-megabyte `?full=1` body the item was written for is still unbounded; what you have covers a server that accepts the connection and never replies. Could you read the body inside the helper and clear the timer after it? All five call sites call `.json()` immediately and two also read `res.headers.get('server-timing')`, so returning `{ data, headers }` covers them. Same scoping applies to `_terminalCaptureInflight`, which decrements at headers, so bodies still streaming do not count toward the budget of a capture starting beside them.

**2. The precache still never gets used (src/web/public/sw.js:56).** `renderIndexHtml` runs `cacheBustAssets`, which appends `?v=<mtime>` to every same-origin `.js`/`.css` reference, content-hashed names included. On my instance that is `src="app.556be563.js?v=1789423735875"`. The precache stores `/app.556be563.js` with no query, and `caches.match(request)` is query-sensitive, so nothing in the list can be matched: only `/`, the icons and `/manifest.json` are reachable, which was already the case. The change therefore turns entries that failed cheaply into entries that succeed expensively, about 1.3 MB uncompressed downloaded a second time at install, and since `CACHE_NAME` now carries the build id that install happens after every deploy. One line fixes it: `.catch(() => caches.match(request, { ignoreSearch: true }))`, which also lets runtime-cached entries survive an mtime change. The build-side plumbing is good and I want to keep it: I simulated the rewrite against a fake manifest and both anchors hit exactly once, the result parses, and `APP_SHELL` resolves correctly at the root and under a `/codeman` mount. The per-build `CACHE_NAME` is a genuine fix for unbounded cache growth.

Two smaller ones while you are in there.

**3. `_wsOutputGapSession` is only cleared in `ws.onopen` (app.js:2956).** Two other paths reload the same buffer without clearing it, so the reconcile runs a second full replay. Switch away from a dropped session and back: `selectSession` reloads the buffer and only then calls `_connectWs` (app.js:6623), so by the time the socket opens neither the `_isLoadingBuffer` nor the `_terminalRefreshOwner` guard applies. And on an SSE reconnect, `handleInit`'s `keepTerminal` branch already calls `_onSessionNeedsRefresh` (app.js:4036-4043) for exactly this reason (its comment says "so output produced during the outage is not lost"). That second point means the PR description overstates the gap: when a phone leaves the network, SSE drops too and its reconnect already reconciles. The hole you are genuinely closing is narrower and still worth closing, a WebSocket dying while SSE stays up, where `_onSSETerminal` keeps discarding SSE terminal frames until `_wsReady` flips false. Clearing the marker whenever a buffer reload for that session completes, and in `_cleanupSessionData`, covers it.

**4. CLAUDE.md:337 says "all of them measured rather than reasoned",** which your own "Not verified here" section contradicts. That line is load-bearing in this file: its value is that a reader can trust a claim without re-deriving it. Please say which half was measured (the replay race) and which is reasoned from xterm's source. You can now describe the watchdog's mechanism as verified against 6.0.0 given the check above; the iOS rAF-discard premise stays reasoned. While editing: the WebSocket reconcile and the sw.js/build.mjs contract get no CLAUDE.md line at all, and the sw contract in particular is the kind of build-time "keep these two in sync or the build throws" rule this file carries elsewhere.

Nits, take or leave: `test/xterm-private-api.test.ts:37` pins the range string from package.json, so a real upgrade inside `^6.0.0` slips through while an innocuous range edit fails, reading the resolved version from package-lock.json would observe the actual dependency; and `test/sw-precache-manifest.test.ts:61` hand-copies the `HASHABLE` list out of scripts/build.mjs, which is the same drift the PR is fixing (the test already reads that file).

One process note for next time: this is six changes in one PR, and the service worker plus build script is a different subsystem from the four terminal fixes, with the only build-breaking risk in the set. Splitting that out would have let the terminal work land on its own timeline.

Everything else is green here: typecheck, lint, format, frontend-syntax, public-assets, and the full gate (386 files, 7273 tests, 546s). The cron-service failure you saw locally does not reproduce on my machine. Send the fixes for 1 and 2 (plus 3 and 4 if you have the appetite) and I will merge. I would still like a real-device pass on the watchdog before the release that carries it, as you suggested.

rounakdatta:
Thank you for this — particularly for verifying the field path and the heal under jsdom. That covers the half I flagged and couldn't reach, and it's the reason I'm comfortable with the watchdog now.

All four are addressed in d200c0e4, plus both nits. Two of them were defects in my previous commit and one was a regression I introduced, so taking each in turn.

**1. Deadline only covered time-to-headers.** You're right, and I reproduced your measurement before fixing it: headers at 30ms, timer cleared there, body completed at 4026ms unaborted under a 1000ms deadline. So the item didn't do the thing it was written for — it only bounded a server that accepts a connection and never replies.

`_fetchTerminalCapture` now reads the body and returns `{json, headers, headersAt}`. `headers` for the two `server-timing` reads; `headersAt` because those same two callers measure header-vs-body time and can no longer observe that moment themselves. `_terminalCaptureInflight` decrements after the body for the reason you gave. Same stalled-body test now aborts at 1005ms.

It has a behavioural test against a real socket, plus a source guard asserting `await res.json()` precedes the `finally` — I checked that guard fails when the helper is reverted to the old shape, since a behavioural test using its own local copy would have passed either way.

**2. Precache unreachable.** Confirmed on the running instance — `vendor/xterm-zerolag-input.6fee72f2.js?v=1789402869101`, hash *and* query. So my change was strictly worse than the status quo: it turned entries that failed cheaply into ~1.3MB downloaded at every install that nothing could read back, once per deploy now that `CACHE_NAME` rotates. `{ ignoreSearch: true }` applied, and pinned by a test since it's subtle enough to be tidied away by someone cleaning up that handler.

**3. Marker only cleared in `ws.onopen`.** Both paths confirmed. `selectSession` loads the buffer and only then calls `_connectWs`, so by the time the socket opens neither guard applies and the reconcile replayed everything a second time. `_markTerminalBufferReconciled()` now runs from `_onSessionNeedsRefresh`'s `finally`, from `selectSession` after its load, and from `_cleanupSessionData`.

Your second point there is the more important one: the PR description was wrong, not just imprecise. SSE drops with the network and `handleInit` already reconciles, so "output produced while a phone is off-network" described a case that was already handled. The genuine hole is the WS dying while SSE stays up — `_onSSETerminal` discards frames while `_wsReady` is true, and that only flips in `onclose`, so a half-open socket costs up to the ping+pong window. I've rewritten both the code comment and the PR description to claim only that.

**4. CLAUDE.md overstated.** Fair, and it's the failure mode that file exists to prevent. Now split explicitly: the replay race is **measured**, the watchdog mechanism is **verified against 6.0.0** (credited to your check), the iOS rAF-discard premise is **reasoned** and still wants a device. Added the two missing entries — the WebSocket reconcile, and the sw.js/build.mjs contract as a build-time "keep these in sync or the build throws" rule.

**Nits, both taken.** The range assertion was wrong in both directions, exactly as you say — 6.4.0 can rename a private field while resolving inside `^6.0.0`, and a harmless range edit failed while changing nothing installed. Reads the resolved lockfile version now. And the hand-copied `HASHABLE` list was the same drift this PR exists to fix, which I should have caught; it's parsed out of `build.mjs`, with a guard so an empty parse can't make the test vacuously pass.

**On the split.** Agreed. The sw/build work is a different subsystem and carries the only build-breaking risk in the set, and separating it would have let the terminal fixes move independently. Happy to split it now if you'd prefer — say the word and I'll put the sw/build commit on its own branch and rebase this one down to the four terminal fixes.

Gate here: 7276 passed, 1 skipped, and the same `cron-service` sensitive-path failure that doesn't reproduce on your machine — environmental, and it fails identically on clean `origin/master` in my sandbox.

Still no browser my end, so the real-device pass on the watchdog is still worth doing before the release that carries it, as you said.

Ark0N:
Thanks, this is a clean second round: all four items from the first review are in, and the two you classed as defects in your own previous commit are fixed in the right place rather than papered over. For the record I reproduced the replay race independently against this repo's xterm 6.0.0 before reading your fix: `write('p8'); reset(); write('rmissions')` renders `p8rmissions`, the queued `\x1bc` renders `rmissions`, and RIS really does return the buffer to its starting length after 30 written lines. Full gate here is 386 files, 7277 tests, 551s, green, and GitHub CI is green on both jobs.

Six small things, none of them blocking. I will apply the first two at merge unless you would rather do it.

1. `src/web/public/app.js:2374` is a terminal capture that still has no deadline. It is the response viewer's byte-buffer fallback, and it uses the no-param `/terminal` form, which `session-routes.ts:2748` caps only at `terminalBufferMaxBytes` (32MB by default), so it is the largest body the frontend asks for anywhere. CLAUDE.md:337 now says "Every terminal capture carries a deadline", and that line has to be true as written. Either route it through `_fetchTerminalCapture` (it reads only `.data`, so the shape fits unchanged) or narrow the sentence.

2. `src/web/public/app.js:2664`: `_onSessionClearTerminal` passes no opts, so it gets `FETCH_DEADLINE_TAIL_MS` (15s) for that same uncapped no-param capture. Largest payload, smallest budget. Passing `{ full: true }` fixes it.

3. `src/web/public/app.js:2993`: `ws.onopen` nulls `_wsOutputGapSession` before it fires the fire-and-forget reconcile, and `_onSessionNeedsRefresh`'s `finally` (app.js:2639) clears it again on every exit including the catch. So a reconcile that fails, or that hits the new abort deadline, drops the gap and nothing retries it, which is the flaky-link case the marker exists for. Clearing only after a repaint actually happened would close it. Happy to take this as a follow-up issue instead if you would rather not reopen the file.

4. `src/web/public/app.js:6469`: for a busy or shell session, `selectSession` clears the pane at 6441 before the fetch, so an aborted capture now leaves a blank terminal, discards queued live output in the catch, and never reaches `_connectWs` at 6672. The error path predates this PR, but the deadline makes it reachable on a link that is merely slow rather than broken. A single `tail=` retry on `AbortError` would degrade it to a smaller capture instead of a dead tab.

5. Nothing in the gate pins the replay-clear invariant itself, which is the drift this PR is fixing. Two asserts in `test/terminal-resilience.test.ts` in the same style as your `_fetchTerminalCapture` guard would cover it: `_resetTerminalForReplay`'s body contains the queued write and no `.reset()`/`.clear()`, and the two replay call sites contain no `this.terminal.clear()`.

6. `src/web/public/terminal-ui.js:3329` still says the version guard "pins the dependency RANGE". After the nit was taken it pins the resolved lockfile version, so any bump fails it, not only a major. CLAUDE.md already says this correctly.

Two genuine nits I will just fix: `app.js:79` hardcodes 300 where `DIAG_ENTRY_MAX_CHARS` exists, and `panels-ui.js:5266` and `5276` still have `clear()+reset()` pairs (not replay paths, so the invariant holds, but worth routing through the same helper).

On the split: leave it. The build rewrite is covered by `test/sw-precache-manifest.test.ts` and everything is green, so separating it now costs a rebase and buys nothing. Worth doing next time, as you said.

Send the fixes for 1 and 2, or tell me to take them at merge, and this goes in. The real-device pass on the watchdog stays on my list for the release that carries it.

rounakdatta:
Rebased onto current master (179 commits; #436's replay-tail cutoff and #431's queued RIS compose cleanly — the two `_fetchTerminalCapture` conflicts resolve to `capture.headersAt`, which is the field that exists for exactly that reason). All six items from your third pass are in, plus both nits you offered to take at merge.

But the headline is that this PR now also fixes **#464**, and I owe you a correction: the four items you reviewed do **not** fix it. I went looking because Rounak asked whether the cause matched, and it does not. Taking your items first, then that.

---

## Your third-pass items

**1 & 2 — the uncapped captures.** Both are the no-param `/terminal` form, so `terminalBufferMaxBytes` (32MB) is their only ceiling. `app.js:2374` had no deadline at all and `_onSessionClearTerminal` had the 15s tail budget for the largest payload the frontend asks for anywhere. Both take `{ full: true }` now, so CLAUDE.md:337 is true as written rather than narrowed.

**3 — the marker cleared before the work it guards.** Both halves confirmed, and the fix is the one you described: `_markTerminalBufferReconciled()` moved out of `_onSessionNeedsRefresh`'s `finally` and onto the repaint itself, right after `_syncStickyScrollBaseline()`. `ws.onopen` no longer nulls it up front either. So a reconcile that throws, is skipped, or hits the new deadline — the flaky link the marker exists for — leaves the marker set and the next open tries again. Re-entry is safe because `_terminalRefreshOwner` already makes a second reconcile for the same session a no-op.

**4 — an aborted capture leaving a dead tab.** A `?full=1` capture that hits its deadline now retries once on `tail=`, and `_fullHistoryLoaded` is un-set so the next select may try the full pull again. Two more things were wrong on that path once I looked: the catch never reached `_connectWs`, so a failed load left the session mute until you switched away and back, and `_setTerminalLoadState(…, 'failed')` only marks the tab — nothing ever cleared it, so a failed select left the spinner and `aria-busy="true"` on forever. The catch now opens the socket, clears the load state, and writes one dim line where the content would have been, but only when the pane was actually blanked; a pane still holding its previous content is stale, not empty, and stacking a notice on readable output is worse.

**5 — the replay-clear invariant.** Pinned, in the style of the `_fetchTerminalCapture` guard: `_resetTerminalForReplay` must contain the queued write and no `.reset()`/`.clear()`, the replay call sites must not hand-roll a clear, and no frontend module may contain a `clear()`-then-`reset()` pair. I checked the last one fails against the panels-ui code it replaced, so it is not vacuous. A lone `clear()` before `showWelcome()` is deliberately still allowed — nothing is written after it, so there is nothing to fuse into.

**6 — the stale comment.** `terminal-ui.js` says "resolved lockfile version" now, and spells out why the range was wrong in both directions.

**Nits, both taken.** `DIAG_ENTRY_MAX_CHARS` replaces the hardcoded 300 — bound through a local first, because `CodemanDiag?.x` still throws a ReferenceError when the identifier was never declared, and that is the one function in the app that must not throw. panels-ui's two kill-all clears route through the helper.

---

## #464 — "text gets muffled sometimes"

Not a dropped frame and not a frozen renderer. It is arithmetic, and the screenshot says so: lines rendered twice, and short tool summaries sitting inside longer prose rows with the prose's tail still visible.

Ink wraps its frame at the width the PTY reported and erases the previous frame by walking the cursor up the rows it *believes* that frame occupied. A browser terminal of a different width makes each logical line take more physical rows than Ink counted, so `eraseLines(n)` clears too few and the new frame paints over rows nothing erased. Reproduced against this repo's own xterm before touching anything — a 120-column PTY against a 62-column terminal renders every wrapped line twice.

Four ways the two drifted apart here, none observable from either end:

1. **`fitAddon.fit()` resizes xterm to the RAW proposal** while every server-facing path reported it floored at 40x10. Measured in Chrome at 430px: font size 44 proposed 13 columns, the server was told 40, xterm stayed at 13. Three call sites each did their own fit-then-floor, and two re-read the proposal *after* the fit — `_shrinkPaddingToFit()` runs exactly there.
2. **`throttledResize` and `sendResize` reflowed locally while withholding the SIGWINCH.** That is the one combination that cannot be right.
3. **The font setters move the cell size and told the server nothing**, so raising the font on a phone left the CLI wrapping at the old column count.
4. **`Session.resize` declines a small-viewport request under a live desktop claim and says nothing**, because resize was write-only.

You will recognise (1) and (4) — `selectSession`'s `captureMatchesRequestedSize` comment already describes both, calls the clamp "a pane permanently bigger than itself", and says repairing the declined case "needs the pane-ownership policy this does not touch". Both are now fixed at the source, and that comment is updated rather than left saying something that is no longer true.

`syncTerminalGeometry()` is the one function that may change the terminal's size: fit, floor and apply in a single step. A test sweeps every module for a bare `fit()` on the main terminal and finds exactly one — the owner's own. This does **not** change who owns the pane: the desktop still wins and `_startMobileResizeRetry` still takes it back on idle.

For (4) the client cannot win, so it is told the truth: both transports answer a resize with `session.ptyCols`/`ptyRows` (`{"t":"zc"}` on the socket, the body of the resize POST) and the client adopts it. A terminal that keeps a shape the PTY refused does not render "too narrow", it renders garbled. Adopting can leave the pane wider than the screen and the container is `overflow: hidden`, so `.pty-oversized` grants horizontal reach for as long as the mismatch lasts — correct-and-reachable beats correct-and-clipped beats garbled. That rule sets both overflow axes and its own `touch-action` because mobile.css loads later with `.terminal-container { overflow: visible; touch-action: none }`, and a bare `overflow-x` would leave overflow-y computing to `auto`.

---

## Verification

Chrome **does** run on my box now, so this round is not reasoned — it is measured.

Driven against a live server at 430x932, with a second desktop-sized client holding the sizing claim:

- xterm's geometry equals what the server was told, at every step
- font size 44 → raw proposal 13x9 → xterm and PTY both 40x10
- the phone's 41x11 request is declined, the PTY reports 198x43, the phone adopts 198x43
- `overflow-x: auto` / `overflow-y: hidden` / `touch-action: pan-x`, 758px reachable to the right, and the terminal's own vertical scrolling still works

15/15. And the control matters more than the passes: I patched the built bundle back to the old fit-raw-report-floored behaviour and re-ran — `{"proposed":{"cols":13},"returned":{"cols":40},"xterm":{"cols":13}}`. Restored: `xterm: 40`. That 3x column mismatch is the bug, in a real browser.

The reproduction test is written as a contrast — the same Ink stream at *matching* widths must render each line exactly once — so it cannot be satisfied by code that fixes nothing.

**Gate**: 419 files, 7890 tests, 705s — 7877 passed, 12 skipped, 1 failed. The `cron-service` "/etc/passwd blocklist" failure is environmental; I confirmed it this time rather than inheriting the claim, by running that file against a clean `origin/master` worktree, where it fails identically.

The gate also caught fourteen failures the focused tests could not — every harness that builds a partial app from cherry-picked mixin methods. Those are a separate commit. One of them, `detached-session-pane-sizing`, pinned the behaviour item (2) reverses, so its assertion is restated with the reason rather than patched.

## Still not verified

The iOS rAF-discard premise behind the watchdog, as before — that still wants the real-device pass. And #464 is verified on desktop Chrome emulating a phone, not on Rounak's actual Android; the mechanism and the fix are measured, the report itself is not reproduced on his hardware.

rounakdatta:
One more pass, pushed as `e587d845` — a third look at the states nobody had actually opened.

**The failed-load notice did not fit.** The line that `selectSession`'s catch writes into a blanked pane was one 70-character sentence. At 430px it exactly filled the row; at 320px it wrapped and left a lone `.` on a line of its own. The narrowest terminal this app will render is the 40-column floor — reachable today just by raising the font on a phone — so it is three lines now, none over 25 columns, one fact each: what failed, that the session is still alive, and what to do. It says *reload* rather than "reopen the tab", because `selectSession` early-returns when the session is already active, so clicking the tab you are already on retries nothing. The first version named no next step at all, which is a mostly-empty terminal and no way out of it.

**CLAUDE.md no longer cites "758px reachable to the right".** That figure is a property of the test content, not of the fix, and this file's value is that a reader can trust a claim without re-deriving it. What it pins instead is the invariant that survives any content: the full pane width is reachable, and removing the class returns `scrollLeft` to 0.

That last one came out of checking a case I had not: scroll the oversized pane right, then let the mismatch resolve. If `scrollLeft` had stayed put, the content would have been parked off-screen with `overflow-x` back to `hidden` and no gesture to bring it back. It resets to 0 — verified, not assumed.

Four other states checked while I was there, all fine: the horizontal scrollbar steals no terminal height, repeated reports produce at most one notice, the desktop that *owns* the pane is never marked oversized, and vertical scrolling still works under `touch-action: pan-x`.

I also re-audited your six items against the code rather than against my own summary of it — all six plus both nits confirmed present.

**One thing I did NOT do, deliberately.** The shared toast has no `role="status"`, so the "another device is setting the width" notice is never announced to a screen reader. Every other transient banner in the app (`offlineBanner`, `hostWakeBanner`, `rebootRestoreBanner`, `historyTruncationBar`) carries it, so the toast looks like the odd one out — but changing it alters how *every* toast in the app behaves, and that is your call rather than something to slip into this PR. Happy to do it here or open it separately, whichever you prefer.

CI green on both jobs for `d9fa9ba1` (typecheck+lint 1m13s, unit+integration 8m37s) — which also settles the `cron-service` question: it passes on CI, so that failure really is only my box.

Ark0N:
Thanks for this, and for how completely each round has been answered. The queued RIS clear, the deadline that actually covers the response body, and the build-generated precache are all things I want in the tree, and the items from my last pass are fixed at the right place rather than worked around. The gate is green here (419 files, 7890 tests, 566s; the one failure was a `timeoutMs: 39999` clock artifact in the remote-wake path, which passes on a solo re-run).

The #464 geometry work landed after that review, so this is its first read, and it needs another pass. Two of the items below I could only see by running it, so I have included what I measured.

**Blockers**

1. **Adopting the PTY's rows puts the CLI's input line off-screen (src/web/public/terminal-ui.js:5594, src/web/public/styles.css:3826).** `.pty-oversized` grants horizontal reach only, on purpose. In headless Chromium against this repo's own CSS and vendored xterm, a 360x300 container adopting 198x43 (your own numbers) paints an `.xterm-screen` 688px tall, `.xterm-viewport` has nothing to scroll (`scrollHeight - clientHeight` is 0), and the last row sits below the container with the user's `scrollTop` pinned at 0. So the phone shows the top 18 of 43 rows and cannot reach the composer: output visible, typing invisible, for as long as the desktop keeps the claim hot (90s idle window, 30s retry). Before the change xterm stayed at 18 rows, so the frame was garbled but the newest row, including the composer, was on screen. Could we adopt columns only? Width is the axis Ink's wrap and `eraseLines` arithmetic depends on, and leaving rows alone keeps the composer at the bottom of a scrolling viewport.

2. **`test/capture-geometry-retry.browser.test.ts:526` now fails.** `expect(requested!.rows).toBeGreaterThan(await terminalRows(page))` gives `expected 10 to be greater than 10`. Its premise is the clamp itself (floored report, unfloored fit), which this PR removes at the source, so it can never hold again at any viewport. It passes 8/8 at the merge base and fails 1/8 on this head. The file is in `BROWSER_TEST_GLOBS`, so both CI jobs stay green while the suite is red: exactly the blind spot CLAUDE.md warns about. The case is still worth keeping; it needs a premise that does not depend on the clamp, and the comment above it needs the correction the `selectSession` comment already got.

**Majors**

3. **A session with no live PTY reports 120x40 and the client adopts it (src/web/routes/session-routes.ts:2123, src/web/routes/ws-routes.ts:254).** `Session.resize()` writes `_ptyCols`/`_ptyRows` only when `this.ptyProcess` is set (src/session.ts:3891), and they are never seeded from the spawn geometry, so a dead-pane session (or one created through the API and not started) answers the constructor defaults. I checked against the real class: a fresh session reports 120x40, swallows a 45x20 mobile resize, and still reports 120x40. Clicking that tab now resizes the browser terminal to 120x40 and, on anything narrower than 120 columns, pops "Another device is setting the width" with no other device involved. Reporting nothing when there is no PTY is enough; `reconcilePtyGeometry` already treats a non-finite report as no evidence.

4. **The 40-column floor clips the right of the pane with no way to reach it (src/web/public/terminal-ui.js:5448).** Applying the floor to xterm is right, but the floored terminal is then wider than `.terminal-container`, which clips (`contain: layout paint`, and `overflow: visible` in mobile.css does not help). `.pty-oversized` is not applied because xterm and the PTY agree. Measured, 360px container: font 14 gives 41 columns and fits; font 18 applies 40 and paints 433px, so 73px is unreachable; font 24 applies 40 and paints 578px, so 218px (38%) is unreachable. `increaseFontSize` goes to 24 and `_refitAfterCellSizeChange` applies it immediately, so it is two taps away on a small phone, and for a shell session nothing was garbled before. Could `.pty-oversized` key on "the applied geometry is wider than the container can show" rather than only on a PTY mismatch? The CSS already does the right thing once the class is on.

5. **"Scroll sideways" does not work on touch for the sessions it targets (src/web/public/terminal-ui.js:5627).** `terminal-ui.js:809` calls `preventDefault()` on a non-passive `touchstart` whenever the tap classifies as `'content'`, which is the normal case for a mouse-tracking TUI at the bottom of the buffer, and the touchmove handler only ever converts vertical travel into `scrollLines`. In Chromium with touch emulation, a 140px horizontal swipe pans a `touch-action: pan-x` scroller to `scrollLeft` 141 without that `preventDefault` and to 0 with it. It does work for shell sessions, while scrolled up, and with a mouse, which is what a desktop-Chrome measurement sees. Either pan the container from the existing touchmove handler while the class is on, or reword the toast.

**Smaller things, happy to take any of these at merge**

- `docs/architecture-invariants.md:139` still says `getTerminalDimensions()` floors while `fitAddon.fit()` does not. The `app.js` comment was updated; this one is the deep-mechanism copy CLAUDE.md points at.
- `src/web/public/terminal-ui.js:1090`: the `throttledResize` HTTP fallback still discards the reply, which is the one path where a declined resize is most likely to go unnoticed.
- `.changeset/pty-geometry-464.md` is the only changeset, and it does not mention the renderer watchdog, the service-worker precache and per-build cache key, or the crash-trail hygiene. The release notes come from that file.
- `test/terminal-pty-geometry.test.ts:28` imports `@xterm/headless`, which only `packages/xterm-zerolag-input` declares. Green through workspace hoisting today.
- `src/web/public/app.js:2731`: the output-gap marker is cleared inside `if (data.terminalBuffer)`, so a session whose capture is empty refetches on every reconnect.
- `e587d845`'s message says a test asserts the failed-load copy in the built asset. I could not find one under `test/`.

**What happens next.** I would like to take the four resilience fixes and the service-worker change as they are; they are ready. For the geometry half, items 1 to 5 above, then one `npm run test:browser` run diffed against master before you push (that suite is where item 2 lives and CI cannot reach it). If splitting is easier than reworking on this branch, put the geometry commits on their own PR and I will merge the rest now.

rounakdatta:
All five, and the six smaller ones, in `e1e7dc5b`. Both blockers were wrong in ways nothing in CI could have told me, and you answered each with the fix — I took all four of your proposals as written.

Two things up front. **You were right that `e587d845` claimed a test that does not exist**: the assertion lived in a throwaway probe I deleted with the other scratch scripts, so the claim was false when I wrote it. There is a real one now, reading the source rather than `dist/` — `dist/` is not committed, and a test that skips when it is absent passes for the wrong reason in CI. And **I can now run `test:browser` here**, which is why I never saw item 2: Playwright's bundled Chromium needs this box's nix `LD_LIBRARY_PATH` *and* `FONTCONFIG_FILE`, and without the latter Skia aborts with `SIGTRAP` on the first webfont. Every number below is measured, not reasoned.

---

## 1 — Adopting rows put the input line off-screen

Columns only, as you proposed. Width is the axis the wrap and `eraseLines` arithmetic depend on; rows only decide how much is on screen, and keeping the local count keeps the composer at the bottom of a viewport that scrolls.

At your geometry — 360x300 container, desktop holding the claim:

```
phone before:  13 rows x 57 cols
desktop:       41 rows x 198 cols
after adopt:   13 rows x 198 cols     <- rows kept, width taken
screen 202px painted into a 210px container; last row inside the box
758px of the width reachable sideways
```

One trade worth naming rather than implying: a row mismatch under a hot claim can still cost a partial repaint until the CLI's next full redraw. That is strictly better than either previous state — garbled-but-reachable, rather than unreachable — and it is the position your proposal puts us in deliberately.

## 2 — `capture-geometry-retry` and the CI blind spot

Its premise *was* the clamp, so it can never hold again. The case survives on its own terms — a pane already drawing at the requested size must not be replayed — and its premise is now the #464 invariant itself: the floored report and the terminal agree. That is a stronger guard than the old one, because the clamp coming back fails it *here* rather than silently restoring the replay loop. The `stubTerminalAtRequestedSize` docblock that repeated the old premise is corrected too. **8/8.**

## 3 — A session with no pane

`Session.ptyGeometry` returns null without a pane; the HTTP route answers `{}` and the socket sends no frame. I deleted the raw `ptyCols`/`ptyRows` getters rather than leave them available to report the defaults again.

Covered behaviourally against the real class in `session-resize-arbitration.test.ts`, the way you checked it, with the contrast so "always null" would fail: a fresh session reports null, still reports null after a resize it could not apply, and a session *with* a pane reports and follows a resize.

## 4 — The floor clipping

Keys on what does not fit now, and **measured** (`.xterm-screen` against the container, on the next frame — the screen takes its width with the render, so a same-tick read returns the size the terminal just left) rather than derived from cell arithmetic. No PTY mismatch involved anywhere:

```
360px phone, no other device:  57 cols @ font 10  ->  40 cols @ font 24
560px painted into 360px; all 200px of the overhang reachable
overflow-x: auto / overflow-y: hidden
```

`.pty-oversized` is renamed **`.term-overflows-x`** — after this the old name describes only one of the two causes.

## 5 — "Scroll sideways" on touch

Panned from the existing touchmove handler, axis locked once per gesture so a diagonal cannot pan and scroll at once. The CSS grants **no** `touch-action`: handing the browser a pan as well would move the pane twice for one finger on the taps where that `preventDefault` does not run, so JS is the sole owner — the same line `touch-action: none` already draws for vertical.

Under real CDP touch dispatch, on an overflowing pane:

```
140px horizontal swipe  ->  scrollLeft 140   (0 before)
buffer viewportY 47 -> 47                    (no side-effect scroll)
vertical swipe          ->  viewportY 47 -> 34
```

---

## Three defects of mine, found while checking rather than by being told

- `canPanHorizontally` first tested `scrollWidth > clientWidth` alone. That is true of a container that is not a scroller, so a sideways swipe would have locked the axis, done nothing, **and** suppressed the vertical scroll it should have been. Gated on the class as well.
- The notice advised scrolling sideways whenever the PTY was wider, including when it still fitted and nothing scrolled. Now gated on measured overflow — and on a comparison against the width the container *would* request, not the one it currently holds, because once adopted those are equal and the question answers itself false while the condition is still true.
- `_syncTerminalOverflowAffordance` could throw out of `document.getElementById` before reaching its try block. It runs off every geometry change, so a cosmetic affordance could have taken the resize down with it.

## The smaller six

All in. `docs/architecture-invariants.md` records what the clamp used to do and why it cannot any more — edited by hand, because that file is outside the Prettier glob and letting Prettier near it rewrote eleven unrelated emphasis markers. `throttledResize`'s HTTP fallback reads the reply (the path where a declined resize is least likely to be noticed, since no socket means no `{"t":"zc"}` either). The changeset now covers the whole release: geometry, the queued replay clear, the watchdog, the body-covering deadline, the output-gap reconcile, the service-worker precache and per-build cache key, and the crash-trail hygiene. `@xterm/headless` is a root devDependency. The output-gap marker clears after any response arrives, not only a non-empty one — a server answering with an empty capture *has* reconciled us.

---

## Verification

**Browser suite, diffed against master, twice — zero regressions.**

| | Test files | Tests failed |
|---|---|---|
| `origin/master` | 9 failed / 4 passed | 16 |
| this branch, run 1 | 3 failed / 10 passed | 7 |
| this branch, run 2 | 2 failed / 11 passed | 6 |

Run 1 showed one apparent regression, `split-pane > shows existing scrollback immediately on connect`. It passes 2/2 in isolation and did not recur in run 2 — a 3-second poll in a suite running fourteen servers serially. The six that remain fail on master too: five in `opencode-resize` (including `shows "Kill Tmux & Claude Code"`, which has no geometry in it — five `expected undefined to be truthy` in a row reads like a pane that never spawned here) and one `inline-rename` case.

⚠️ **Read that table with care: master is worse on this box than on yours.** You have `capture-geometry-retry` at 8/8 on the merge base; here the whole file fails on master. So several of the entries that "improve" are this machine, not my branch, and I am not claiming them.

**Consolidated live-browser run: 17/17**, covering all of the above plus the invariant, the desktop that owns the pane never being marked overflowing, and no uncaught page errors. Two probes failed on the first pass and both were the probe: the trap check "resolved" the mismatch by reporting the current wide width, which correctly no longer reads as a resolution; and the dead-pane probe deleted the session and got a 404, which proves nothing — that check became the unit test above. Reporting the width the container actually wants does resolve it: 198 -> 69 columns, `scrollLeft` 758 -> 0.

**Gate**: 419 files, 7898 tests — 7885 passed, 12 skipped, 1 failed. The `cron-service` `/etc/passwd` failure is this box: it fails identically on a clean `origin/master` worktree and passes on GitHub CI. typecheck, lint, format, frontend-syntax, public-assets and the lockfile check are all clean.

## Still not verified

The iOS rAF-discard premise behind the watchdog, unchanged. And #464 itself is measured in desktop Chrome emulating a phone, not on Rounak's Android — the mechanism and the fix are measured, the original report is not reproduced on his hardware.

## On the split

The offer stands from my side: if you would still rather take the four resilience fixes and the service-worker change now, say so and I will put the geometry commits (`abf1d1f1`, `d9fa9ba1`, `e587d845`, `e1e7dc5b`) on their own branch and rebase this one down. Everything above is on this branch as it is, so it costs you nothing to decide either way.

rounakdatta:
Opened #470 as the follow-up I offered above, kept off this branch so it does not add a seventh concern to a PR you have twice asked to be smaller.

It is the one remaining live-path risk behind #464, and it is your own principle in a place it had not been applied: the 128KB drop recovery was a fire-and-forget timer that nulled its handle and then called `_onSessionNeedsRefresh()`, whose early returns — a buffer load in flight, a refresh already owning the session — are most likely to be true during exactly the output burst that caused the drop. Skipped recovery, nothing left to retry it, bytes never replayed.

Worth saying what it is *not*: the other two paths I had flagged, `stripInkRedrawBloat` and `formatPaneSnapshot`'s missing erase-to-EOL, are both **capture-only** — the live path is `emit('terminal', data)` straight to the socket and touches neither — so neither can explain corruption in live output, which is what the #464 screenshot shows. I left them rather than fix things that cannot cause the reported bug.

#470 is stacked on this branch, so its commit list carries these four until this merges; the reviewable surface there is its single commit. Nothing in it touches anything here.

Ark0N:
Merged, thanks @rounakdatta! This ships in 1.32.1.

Four rounds, and every one answered completely. The queued RIS clear, the deadline that covers the response body, the renderer watchdog and the build-generated precache are all in the tree now, and #464's "muffled text" finally has an explanation that is arithmetic rather than luck.

The bot's last pass found one real issue, which I fixed at merge: after `_onPtyGeometryReport` adopted a refused width, `sendResize` and `throttledResize` still called `syncTerminalGeometry()` first, so every 30 s mobile retry re-fitted xterm to the container and then back to the PTY (two full re-wraps), and the adopted columns made the next resize run the scrollback clear for a redraw that never came. A new `_geometryForResizeRequest()` keeps the PTY's columns locally while the refusal holds (rows still follow the container) and keeps asking the server for the container width; with no refusal it is `syncTerminalGeometry()` exactly as before. Two tests drive the real methods and are red on your head. Also: `_ptyCols`/`_ptyRows` are seeded at attach and spawn, `terminalGeometryAgrees()` is gone, the stale comment and the xterm-private-api header are fixed, and the fit sweep also matches `fitAddon?.fit?.()`.

I have only verified the watchdog and the geometry path in headless browsers, so if you see anything odd on a real phone, please shout.

PR #441, opened 2026-09-16
fix(input): deliver a recovered keystroke before the Enter that submits it

shenlvkang-collab:
## The report

Every message typed on an Android phone arrives at the CLI with its **last character missing** — `帮我看看代码` is submitted as `帮我看看代`. Reproducible on every send, in both Latin and CJK text, with the local-echo composer on (the mobile default, `cjkInputEnabled: false`).

## Root cause

An Android soft keyboard commits the last typed character and sends the Enter key in **one InputConnection transaction**, so the committed-text `input` event and the Enter keydown are both processed before any zero-delay timer runs.

The orphaned-input recovery added in #388 resolves its candidate only on such a timer, and that loses the character twice over:

1. **Order** — xterm emits `\r` synchronously from the Enter keydown, and `terminal-ui.js`'s local-echo branch submits `pendingText` right there. The recovered character arrived one macrotask too late to be part of the prompt.
2. **Loss** — that same `\r` bumps `canonicalCount`, so by the time the candidate resolved, `canonicalCount > candidate.snapshot` read as *"xterm spoke for this keystroke"* and stood the recovery down. The character was not merely late, it was dropped outright.

The file already guards the counter against xterm's *self-generated* output (DA/DSR/CPR/mouse reports) for exactly this reason, but a genuinely different **keystroke's** canonical data — the Enter — still falsely answered the question asked about the previous one.

## The fix

Drain pending candidates synchronously at the next keydown, from xterm's custom key handler, which runs *before* xterm processes that key. The counter then still holds the value it had while the candidate's own keystroke was current, so the stand-down decision is made against the right keystroke, and the recovered byte reaches the composer ahead of whatever the new key emits. The zero-delay timer stays as the fallback for a keystroke with no key after it.

Physical keyboards are unaffected: there the timer has already resolved the candidate long before the next key arrives.

## Tests

`test/terminal-keycode229-recovery.test.ts` gains two cases, both failing before the change:

- **defect 4** — after `keydown → refused insertText → Enter keydown`, the character is already emitted at the Enter keydown, exactly once, and no timer is left pending.
- a companion case pinning that the synchronous drain does **not** become a "forward everything" path: a keystroke xterm delivered itself is still stood down at the next keydown.

`npm run test:ci` is green (7252 passed). `test/terminal-keycode229-recovery.browser.test.ts` does not start in my environment either before or after this change (the Playwright harness times out waiting for `app.terminal`), so it is unverified here rather than regressed.


Ark0N:
Merged, thanks. This fixes a real defect: an Android soft keyboard that commits the last character and sends Enter in one InputConnection transaction lost that character outright, and draining pending candidates at the next keydown, from inside xterm's custom key handler, is the right place to fix both halves (the ordering, and the counter reading a different keystroke's `\r` as its own).

I reproduced it against the real xterm in headless chromium before merging, driving your exact shape (characters committed one per task, then the last character and Enter dispatched in one task) and recording the bytes that reach the send path. On master the wire is `"hell\r"`; on your head it is `"hello\r"`. Your unit case fails on master and passes on head, the full gate is green here, and `test/terminal-keycode229-recovery.browser.test.ts` does start on this machine, all 4 cases green, so the keyCode-229 self-rescue case is not regressed.

**What I added on the way in** (`ea5323d9`):

A browser case for the batched shape. The unit harness proves which candidate gets forwarded; the ordering is the half that shipped the bug, and only a real xterm shows it. The new case dispatches the character's keydown, its composed insertText and Enter's keydown in one page task and asserts what reaches the send path. I ran it in both directions: with your drain the wire is `o\r`, and with the drain removed it is `\r`, so the character is not merely late there, it is gone entirely. That is worth stating precisely, because it is stronger than "one character can arrive out of order": by the time the zero-delay timer runs, xterm has emitted the `\r` and bumped the canonical counter past the candidate's snapshot, so the candidate stands down and the character is dropped. The other four cases pass in both states.

CLAUDE.md now names the decision point, what it costs and why that cost is acceptable, plus the fact that the pin lives in a suite the CI gate does not run.

**Two notes, neither of which changed the merge.**

Deciding at the next keydown decides with less evidence than the timer did, because xterm's own keyCode-229 rescue (`CompositionHelper._handleAnyTextareaChanges`, a zero-delay diff of `textarea.value` scheduled at keydown) has not run yet. For Enter this is safe: xterm clears the textarea on CR, so its pending diff emits nothing, and I measured no duplicate in any Enter case. Where it shows is two character keydowns inside one task: master sends `"abb\r"`, your head sends `"aabb\r"`. That path is already wrong on master and I could not produce it from a realistic Android input shape, so I took the PR as is, but the trade is worth having written down.

One correction for the record: only "defect 4" fails before the change. The companion case passes on master too, because without the drain that candidate stands down on its timer for the same reason. It is still a useful guard, just not a regression test.

Separately, and not yours to fix: on a device whose keydowns report keyCode 229, xterm emits a stray DEL in this same batched-Enter path (its zero-delay diff sees the textarea that the Enter just cleared), and with local echo on that DEL lands before the `\r` and eats a character in the composer. It is on master too, and your change strictly improves that path (master loses two characters there, your branch loses one). Tracked separately so nobody reads a lingering report as a regression from this PR.

Thanks again, the root cause writeup was correct in both halves and saved me the diagnosis. This ships in 1.30.0.

Ark0N:
Tracked as #448, so the stray-DEL case has a home and nobody reads a later Android report as a regression from this PR.

Issue #411, opened 2026-09-14
feat(session): bring sessions back after a host reboot, not just after a server restart


irisitymichaelgrundberg:
## The problem

Codeman sessions survive a server restart because the tmux server outlives it. They do not survive a reboot of the host, because tmux dies with the machine. Picking yesterday's work back up means finding each conversation in history and resuming it by hand, one session at a time.

The docs already promise something close enough to be read as this. Q&A discussion #293, "Do my agents keep running when I close the browser or my laptop?", answers that `codeman service install` "also survives reboots". That is true of the server but not of the sessions. The systemd unit makes the same assumption explicit: "Agents keep running in tmux when the server restarts, so only signal the server itself" (`src/service-installer.ts`).

Here is what the first boot after a reboot actually does:

1. `reconcileSessions()` (`src/tmux-manager.ts:2572`) lists panes on the Codeman socket and finds none.
2. Every tracked session lands in `dead`, leaves `this.sessions`, and drops out of `mux-sessions.json`.
3. `finalizeRestoredState()` runs `cleanupStaleSessions()` (`src/web/server.ts:2098`), which prunes those sessions from `state.json`.
4. The board comes up empty, and every conversation has to be resumed by hand.

## I prototyped this before filing, so the numbers are measured

The core is about 140 lines of production code. A prototype against `master` at `a017e9a8` adds 536 lines across four files: 135 in `src/web/server.ts`, 5 in `src/web/schemas.ts`, a new 155-line module holding the decision logic as pure functions, and 242 lines of tests. Typecheck, lint, `format:check` and the full `npm test` suite pass, including 19 new tests.

Three things in the existing code make it that small.

**The construction path already exists.** `restoreMuxSessions()` (`src/web/server.ts:2791`) rebuilds a `Session` from `state.json` for each surviving mux session, with the owner re-resolved, the permission mode recomputed through `resolveClaudeModeForUsername`, and env overrides recovered. Restoring a reboot-killed session is that same block with the mux session omitted, so a pane gets created rather than attached.

**Pane creation is not in the route handler.** `startInteractive()` (`src/session.ts:1942`) calls `_setupOrAttachMuxSession()` (`src/session.ts:1593`), which calls `mux.createSession()` whenever the `Session` carries no mux session. `POST /api/sessions` (`src/web/routes/session-routes.ts:873`) never creates a pane at all, so nothing about it needs extracting or refactoring.

[...]

PR #442, opened 2026-09-16
feat(sessions): offer to rebuild the sessions a host reboot destroyed

irisitymichaelgrundberg:
Closes the design agreed in #411.

A host reboot takes the tmux server down with it, so every pane dies, reconciliation finds nothing to attach to, and the board comes up empty. This makes Codeman work out what the reboot killed and offer it back.

## What it does

The boot pass builds a plan inside `restoreMuxSessions()`, in the window where reconciliation has reported the dead sessions and `cleanupStaleSessions()` has not pruned their records yet. That window is the only place the records can still be read, which is why the plan is built at boot even though nothing is rebuilt until a click.

The board then shows a banner — "Restore N sessions from before the reboot" — with Restore and Dismiss. Nothing creates a pane until the user clicks.

## The decisions from the issue thread

- **A click, not an automatic restore.** The heuristic decides whether to ask, never whether to act, so a wrong yes costs a line of text rather than N CLI processes nobody asked for.
- **No setting.** No `settings.json` key, no Zod change, no checkbox, no migration. An always-on variant for headless machines can reuse the same pass later.
- **Respawn and Ralph never re-arm.** A restored session comes back attached, idle and disarmed.
- **`clampEnvOverridesForOwner` moved to `src/session-env-clamp.ts`**, exporting `ownerClampedEnvKeys()` and the clamp. The `_clampEnvOverridesForOwner` test hook stays re-exported from `session-routes.ts`. That module's header records that the reboot path cannot strip anything today, because the persisted override set and claude's privileged keys do not overlap; the grant re-resolution that does bite there is `resolveClaudeModeForUsername`.
- **A vanished workspace is skipped**, at plan time for the banner's count and again at click time for what actually gets a pane.
- **Workspace hooks are installed by the restore route.** `ensureHooksForRecoveredWorkspaces()` and `startStatsCollection` both sit in boot code that has finished by the time the click lands, so the route does both itself. `startStatsCollection` is called unconditionally, since it clears and re-arms its own interval.
- **Scrollback is not restored, and the banner says so.** The pane is new, so the conversation continues and the terminal history does not.

## Which sessions are eligible

A session the user ended deliberately is never revived, and each shape of "ended" has to be recognised from the record alone, because the in-memory guard does not survive a reboot. An unpinned kill has already deleted the record. A pinned kill was demoted to `status: 'stopped'`. A record with no attach process describes a session that never started, or whose pane died outright.

Remote and docker sessions are skipped, since both need another host or a container to be up. So are sessions whose crash-loop breaker had tripped.

**One case this cannot recognise yet.** A session whose agent the user ended with `/exit` IS restored. The pid a record carries is its `tmux attach-session` process, and `remain-on-exit` keeps that pane alive when the agent exits inside it, so Codeman's PTY never exits, no exit handler runs, and the record keeps both its pid and `status: 'idle'`. Nothing durable distinguishes it from a session that was simply idle when the power went. Two real reboots established this, the second after a fix built on the wrong mechanism. #446 covers making Codeman notice the dead pane; until a record can say the agent is gone, this pass offers those sessions back and the user dismisses or closes them.

## Where the pending plan lives

In memory, in `src/web/reboot-restore-registry.ts`, in the style of `web/approval-inbox.ts`. A server restart drops it. That costs the convenience this feature adds and never the conversation, because the conversation is the transcript under `~/.claude/projects` that the Resume list and the Session Manager already read. The module header says so.

An entry a restore is holding stays parked in the registry rather than being forgotten, so a Dismiss arriving mid-restore still reaches it. Parking the entries rather than versioning them is what lets the same ownership predicate apply in both places: the caller spending an entry need not be its owner, since an admin may restore another user's sessions.

## Click-time rules

Hours can pass between the boot that builds the plan and the click that spends it, so nothing is trusted from plan time:

1. The workspace must still be on disk.
2. In multi-user mode the workspace is re-confined to **the entry owner's** current grant, through `isWorkingDirAllowedForUsername`, the way `cron/cron-service.ts` re-resolves a grant at fire time. An admin restoring another user's session is held to that user's confinement.
3. The conversation must not already be live, matched on both session id and conversation id, because the user may have resumed it by hand from the Resume list. Two panes running `--resume` on one conversation is the failure this prevents.
4. Capacity is re-checked on every iteration, because the loop is itself creating the sessions it counts, and the offer can be a day old.
5. Entries leave the plan synchronously before the first `await`, and the route is single-flighted per owner, so a double-click or two devices cannot both reach the same entry.

A rebuild that fails after the session is registered is undone by `discardPartiallyBuiltSession()`, which reverses `registerSessionWithLayout()` and `setupSessionListeners()` and nothing else. It is deliberately not `cleanupSession()`: that is the user-initiated delete, and it would bank the session's historical tokens into the lifetime totals, demote a pinned record to the marker that means "killed on purpose", drop the stored Ralph state and remove the workspace's `.claude-images`.

Ownership scoping runs through `canAccessOwned`, so a user sees and restores their own entries and an admin sees all.

## API

- `GET /api/reboot-restore` — what is on offer, ownership-scoped. The persisted record never reaches the browser.
- `POST /api/reboot-restore/restore` — rebuild all of it, or the ids in `sessionIds`. Each skipped entry carries its own reason, so the banner can tell a missing workspace from a session limit from an agent that would not start.
- `POST /api/reboot-restore/dismiss` — drop the offer.

All three are documented in `docs/api-reference.md`. Each rebuilt session broadcasts `session:created` through the same path `POST /api/sessions` uses, and the banner re-reads the offer on every SSE init, so a board left open across the reboot learns about it without a manual reload.

## What a restored session carries

The `Session` constructor takes only part of a persisted record, and `toState()` is written wholesale, so a naive rebuild would replace the fuller record with a reduced one. `reapplyPersistedSessionState()` restores the rest in two phases. Before the pane spawns it applies what shapes it: the custom-model selection, the nice priority, and the image-watcher flag, which `setupSessionListeners()` reads to decide whether to start a watcher. Afterwards it applies the session's own history: the pin and its original timestamp, token and cost totals, auto-compact, auto-clear, auto-resume, the tab colour and the flicker filter. The history half runs after the spawn on purpose, so that totals never land on a session whose pane failed to start.

The pin matters most. `cleanupSessionsByIds()` keeps a record only while it is pinned, so dropping the pin would hand the record to the next stale sweep.

## Not in this PR

Claude sessions only. The other CLIs name their thread in their own config object, which this does not thread through yet, and that is the natural second PR. Remote and docker sessions are skipped on purpose.

Ralph's loop configuration does not survive a restore: `toState()` reads it off a live tracker, and there is no way to keep it without arming the loop, which is exactly what must not happen automatically. `src/reboot-restore.ts` also records that `os.uptime()` reports the host's uptime, so the boot test never passes for a Codeman running in a container and the feature is effectively off there.

## Testing

`npm run typecheck`, `npm run lint`, `npm run format:check` and the full `npm test` suite all pass: 390 files and 7342 tests, of which 64 across five files belong to this change.

`test/reboot-restore.test.ts` covers the reboot heuristic, every eligibility rule, the workspace check, the already-live dedupe and the plan registry, and drives a real `Session` to prove the construction path creates a resumed pane. `test/routes/reboot-restore-routes.test.ts` covers taking, scoping, single-flighting and re-checking. `test/routes/reboot-restore-rebuild-failure.test.ts` mocks the `Session` module so the route's own construction path runs and then fails, which is the only way to reach the discard. `test/discard-partially-built-session.test.ts` drives the real `WebServer`, because the route tests see only a mock. `test/mocks/mock-route-context-completeness.test.ts` compares that mock against `createRouteContext()` at runtime — `tsconfig.json` includes only `src/**`, so no type check ever compares the two, and the mock went stale twice during this work.

Beyond the suite, this has been run. An isolated instance (`CODEMAN_INSTANCE`, its own tmux socket and port) reproduced a reboot's effects with real agent panes, and the whole path worked: detection, the offer, the click, real panes on the original session ids, and the banner rendered and clicked in a browser at desktop and phone widths. It has since been validated by three actual reboots on a live board.

Three claims the code used to make did not survive contact with a real server, and are corrected rather than left standing. A missing agent binary does not make `startInteractive()` throw, because the resolver finds it by absolute path. An unenterable workspace does not either, because tmux falls back to another directory. And `/exit` does not persist a null pid, for the reason given under eligibility above — a rule written on that assumption was shipped, disproved by the next reboot, and its documentation corrected in place.


Ark0N:
Merged, thanks, and thanks for the three real reboots behind it. It works out what a host reboot destroyed and offers those sessions back behind a click, without ever spawning anything on a guess, which is the shape we agreed in #411.

I fixed the small things on the way in rather than sending it back (`bb8ada7e`). Listing them so you know what changed and can push back if I have any of it wrong.

**1. The rebuilt `Session` did not get `nameSource`**, so the constructor re-inferred it from the name. A session the user renamed by hand to something that matches `w<n>-<case>` came back as `placeholder`, and with auto-naming on (which shipped in 1.29.1, after you branched) the next prompt overwrote their name. The route persists right below, so the loss went to disk. Now passed through the way `restoreMuxSessions()` already does.

**2. `liveSessionIds` and `liveConversationIds` were snapshotted once** before a loop that awaits a real `startInteractive()` per entry, while capacity and the workspace check were re-evaluated per entry. By the tenth entry that snapshot is tens of seconds old, and a conversation resumed by hand from the Resume list during the restore was invisible to it: two panes on one transcript, which is the thing the check exists to prevent. Both are now read per iteration, and the late case is spent rather than re-offered, for the same reason the batch case is: unlike a missing workspace or a withdrawn grant, an open conversation is not a condition that stops being true.

**3. The auto-resume question, which I said I wanted your view on.** I took the drop-the-stamp option rather than hold the PR for a round trip, because the argument only goes one way once it is written out: the stamp predates the reboot, the pane is new, and one click on eight sessions means eight unattended `continue` prompts about a minute later, against the module header's own promise that a restored session comes back attached, idle and disarmed. So `reapplyPersistedSessionState` gained a `rearmAutoResumeSchedule` option, default true, and this path passes false. Auto-resume stays **enabled**, so it re-arms on the next real limit message; only the stale schedule is dropped. A Codeman restart still re-arms from the stamp, because there the limit footer will not reprint on its own and dropping it would strand the pause. Say the word if you would rather it went the other way and I will flip it.

**4. The container claim in the heuristic header** was too pessimistic in one direction and too quiet in the other. After a genuine host reboot the container sees the host's short uptime, so the banner does appear and the feature works. What does not work is a container-only restart, and since `docker/server.Dockerfile` installs tmux inside the Codeman container and the self-updater restarts the Compose deployment by exiting the container, that is the case where this would help most. Comment corrected; closing the gap needs a boot signal the container owns (PID 1's start time, gated on the existing `isRunningInContainer()`), which is a follow-up rather than something to widen the heuristic for.

**5. `docs/api-reference.md`.** 17 of the 18 hunks were Prettier reformatting of prose the PR does not otherwise touch, and `docs/` is outside the format glob in `package.json`, so this looks like editor format-on-save leaking in. Worth flagging that it was not purely cosmetic: it also de-indented continuation lines in two existing bullets, and in your own new section. Reverted to the hunk that adds the Reboot restore section, with those continuation lines repaired.

**Picked up with no action needed from you:** a route test for the multi-user `workspace-forbidden` branch (the only new rule with nothing behind it, and the one where getting the owner-versus-caller distinction wrong would be invisible); the CLAUDE.md entry, plus the route-file and frontend-module counts that go stale with it; `@loadorder` renumbered to 11.65, since 11.7 is `admin-ui.js`, which loads after you; the banner hidden under `body.solo-mode`, since a detached single-session popup was offering to restore sessions it has no tab strip to show; the two teardown steps the discard omitted that `_doCleanupSession` performs (`recordSessionStopped()` and `ralphTracker.fullReset()`, both cosmetic); and `seedAgentSessionPreamble()` on the restore path, which both create paths call.

**Everything else holds up.** Typecheck, lint, format, frontend syntax and the full `npm test` gate all pass, and the two paths that could actually do damage, the failure undo and the ownership scoping, are the two with the most tests behind them. Writing the `/exit` limitation into the module header instead of papering over it is the right call, and `discardPartiallyBuiltSession()` spelling out every reason it is not `cleanupSession()` is the kind of comment that stops the next person undoing it. Good work. This ships in 1.30.0.

irisitymichaelgrundberg:
Thanks for merging, and for fixing these rather than sending them back.

**Auto-resume: keep it as you have it.** The argument is stronger than you put it. The stamp is a wall-clock time derived from a limit message the old pane saw, and the rebuilt pane is a different process with no evidence it is still limited. If the reboot happened after the stamp, the schedule is not "a prompt a minute later" but already overdue, so it fires almost immediately — several unattended prompts right after a click whose whole promise is idle sessions. The reason the ordinary restart path re-arms does not carry over either: it re-arms because nobody is present and the limit footer will not reprint, whereas a restore happens because someone just clicked. Leaving auto-resume enabled and dropping only the stale schedule costs nothing, since the next real limit re-arms it.

**The already-live snapshot is the one I should have caught.** Re-reading it, the mistake is visible in the shape: I re-evaluated capacity and the workspace per entry, for exactly the reason that the loop changes the world as it runs, and then read the live sets once outside it. Four review passes and three real reboots went by without finding it, because nothing ever exercised a slow multi-entry restore against a board someone was using at the same time. Reading both per iteration is right, and so is spending the late case rather than re-offering it.

`nameSource` is a fair catch, and the auto-naming change is what turns it from cosmetic into data loss — that landed after I branched and I did not go looking. The container caveat I simply got wrong in both directions, and gating a PID 1 start time on `isRunningInContainer()` sounds like the right shape for the follow-up. The `docs/` churn was editor format-on-save leaking past the glob; sorry for the noise, and thank you for repairing the continuation lines rather than just reverting.

Happy to take the other CLIs' resume ids as the next one, unless you would rather that waited for #446 to settle what an exited session's record looks like.

PR #453, opened 2026-09-19
feat: split-pane sessions — view two live terminals side by side

timkjr:
## Summary
- Adds a "Split" header button (opt-in, `showSplitButton`, off by default) that opens a picker to view a second session live, side-by-side with the active one, via a draggable divider.
- New `SplitTerminalPane` class (`terminal-split.js`): its own xterm instance + WebSocket to `/ws/sessions/:id/terminal`, independent of the primary pane. Deliberately plainer than the primary pane (no local-echo overlay, CJK IME, touch handlers, or keyboard accessory bar) — see the design discussion for why.
- Handles both session-ends-while-split edge cases: Pane B's session ending collapses back to Pane A at full width; Pane A's session ending promotes Pane B to the new primary pane.
- No persistence by design (v1 scope) — a page reload always returns to single-pane.

Design discussion: #452

## Fixed during implementation (not in the original design)
- Pane B was blank on connect until the target session happened to produce new output, because the WebSocket route is subscribe-only (no buffer replay on connect). Fixed by fetching existing scrollback (`?full=1`) before opening the socket.
- Neither entering a split nor dragging the divider told Pane A's own session about its new size — only a passive, debounced `ResizeObserver` did, eventually. Both now force a resize immediately (drag end only, not per-move, to avoid flooding the PTY).

## Test plan
- [x] `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run check:frontend-syntax`
- [x] `npm test` (full suite, 7393 passed)
- [x] `npm run test:browser` for the split-pane suite (7 tests: connect/echo/destroy lifecycle, scrollback replay on connect, open/close reparenting, split-picker filtering, force-resize on split-open and drag-end)
- [x] Manually verified live: split/un-split/re-split cycling, divider drag both directions, Pane A/B redraw correctness

https://claude.ai/code/session_012AcpEgRwdEbdTqm4W4wanS


Ark0N:
Thanks for this, and for the amount of care that clearly went into it. Opening a second live session beside the active one, with its own xterm and its own WebSocket and a divider you can drag, is a real feature and the core of it is well built: patching `_onSessionDeleted` and `selectSession` on the prototype at module-evaluation time is exactly right (`connectSSE()` caches the handler reference on first connect, so nothing later would have fired), the picker escapes everything it interpolates, and `test/split-pane-hidden-button-css.test.ts` is a guard this repo will keep whatever happens to the rest. All your checks reproduce here: full gate 7393 passed, and all 7 of the browser tests pass on my machine too.

Before this can land I need a few things.

**Scope.** The branch also carries four changes that have nothing to do with split-pane and are not in the description: the sticky-scroll throttle (`src/web/public/terminal-ui.js:3388` and `:3773`), Ctrl+Backspace sending Ctrl+W (`terminal-ui.js:433`), the prompt-preview line in the welcome Resume list (`terminal-ui.js:2385`) and the removal of the phantom `testcase` picker entry (`session-ui.js:67`). Please take them out of this PR. The sticky-scroll one in particular needs to stand on its own: it changes the write path for every session and every CLI on the strength of one upstream omp bug, and that path has a long comment about how its two viewport guards interact. The `testcase` removal is a genuine fix and I am happy to take it as its own small PR.

**DECISIONS.md.** Please revert the `.gitignore` change and drop the file. The entry you removed says "In-session decision scratchpad (context-survival mechanism, not a deliverable)", and the committed file contains your own deployment log (systemd override repoints, a `tlkMods` branch that does not exist here). Un-ignoring the path also means every contributor's scratchpad starts showing up as untracked. `.worktrees/` in `.gitignore` is fine, keep that.

**Pane B never re-fits when the window changes size** (`src/web/public/terminal-split.js:112`). `fit()` is only ever called from the divider drag, and `throttledResize` in `terminal-ui.js:1020` only touches Pane A. I checked this in a real browser: split at a 1600px viewport, shrink to 800px, wait for the debounced observer, and Pane A's cols change while Pane B stays at 98, both in xterm and in its PTY. Same for the Alt+B sidebar toggle and the tab-rail drag. Please observe Pane B's container, or call `this._splitPane?.fit()` from the existing trailing edge.

**Pane B silently eats keystrokes after a dropped socket** (`terminal-split.js:47` and `:103`). `onData` discards the keystroke when the socket is not OPEN, and there is no reconnect, so after a server restart or a tunnel blip the pane shows stale content and swallows everything typed into it with nothing on screen to say so. Also: `{t:'i', d}` with no `cid`/`seq` is the frame `app.js:3114` reserves for scroll ticks, whose own comment says "a dropped scroll tick is harmless; keystrokes still go through `_sendInputAsync`". A visible dead state in the Pane B header (or auto-collapsing the split) would be enough for v1.

**Splitting against a popped-out session fights the popup for its PTY size** (`terminal-split.js:118`). `sendResize` deliberately yields for a detached session ("One PTY cannot hold two sizes"), and `_sendResize()` has no such check. A detached session still sits in `this.sessions`, so it shows up in the picker. The simplest fix is to exclude `detachedSessions` from `buildSplitPickerSessions`.

**Nothing gates the feature to desktop.** The spec says it is gated the way `home-sessions.js`'s rail is, and that module has both a JS width check and a media-query backstop. Here there is neither, and `showSplitButton` is a synced setting, so turning it on at a desk also puts the button in the phone header. On a 393px phone the two 240px min-widths plus the divider need 486px, so Pane B lands off-screen, and the divider has no touch handlers. Either add the gate, or move `showSplitButton` into the `displayKeys` per-device set in `settings-ui.js:3148`.

Smaller things you could fold into the same push: `terminal-split.js:362` should pass `{ auto: true }` to `selectSession` (an app-driven selection must not spend the session's idle alert, see the Approvals Inbox note in CLAUDE.md); `openSplitPane` should return early when there is no active session, otherwise a split opened from the home screen is created and connected behind the opaque welcome overlay; and `onMove` at `:316` needs an `if (!container) return;` or it throws on every mousemove when the split auto-collapses mid-drag.

One more thing worth settling before you spend more evenings: #452 went up the same day as this PR and nobody has replied to it yet. Let me answer there on whether I want an in-page divider alongside the existing pop-out window, so the design question is closed before the code question. Docs-wise, if it goes ahead, I would keep the design spec as `docs/split-pane-sessions-plan.md` alongside the other `*-plan.md` files and drop the 1106-line task plan, which is workflow scaffolding rather than repo documentation.

Once the unrelated commits are out and those five are addressed I will take another pass. Nice work for a first contribution.

Ark0N:
Thanks for turning the last review around so quickly, and for the care that went into it. This now does what the description says: a second live session beside the active one, its own xterm and WebSocket, a draggable divider, and a clean collapse when either session ends. All six items from my previous pass are addressed, the unrelated commits are gone, and the full gate is green here too (408 files, 7732 tests, 569s).

One blocker and two majors before it can land.

## Blocker: the new setting breaks every App Settings save

`src/web/public/settings-ui.js:2124`

`saveAppSettings()` puts `showSplitButton` in the body it PUTs to `/api/settings`, but `SettingsUpdateSchema` is `.strict()` and does not declare it, so the server answers `400 INVALID_INPUT` and persists nothing at all. `_api()` returns the Response instead of throwing and the call site never checks `res.ok`, so the UI still shows "Settings saved" while `~/.codeman/settings.json` is untouched. That takes down `workspaceHooksEnabled`, `agentSkillEnabled`, `tunnelEnabled`, `claudeModel`, the three watcher toggles, notification preferences and voice settings, on every save, on every device.

I reproduced it with `app.inject()`: the same body with the key gets `400 {"success":false,"error":"Invalid settings","errorCode":"INVALID_INPUT"}`, without it 200.

Moving the key into `displayKeys` was half of what I asked for. The other half is the strip: `displayKeys` is a client-side merge policy, not a wire filter. Add `showSplitButton: _ssp,` to the destructure at `settings-ui.js:2346-2383`, right next to `showFileViewerButton: _fvb,`, which is the setting your invariants entry says this one follows. `test/terminal-auto-copy.test.ts:315` is the guard for the three-way rule (in `displayKeys`, stripped from the PUT, absent from the schema); a copy of it for this setting would stop it recurring.

## Major: the divider drag is unthrottled

`src/web/public/terminal-split.js:326`

`onMove` calls `this.fitAddon.fit()` and `this._splitPane.fit()` on every mousemove, and `_splitPane.fit()` sends a `{t:'z'}` frame each time with no unchanged-dimensions skip. Every changed column reaches `Session.resize()`, which spawns a fire-and-forget `tmux resize-window` child plus a SIGWINCH, so one drag across half a wide viewport is roughly fifty of each into Pane B's CLI. Pane A gets a full xterm reflow per move too, which is exactly what the comment on `throttledResize` (`terminal-ui.js:998`) exists to prevent, and your own design spec says the drag is rAF-throttled.

Keep the `flexBasis` writes per move, rAF-throttle the two `fit()` calls, and send Pane B's resize from `onUp` where Pane A's already is.

## Major: Pane B pulls the whole scrollback, unchunked

`src/web/public/terminal-split.js:65`

It requests `?full=1` for every mode and writes the body with a single `terminal.write()`. The primary pane deliberately does neither: `app.js:6445` sends `?full=1` only for non-shell sessions and falls back to a 1 MiB `?tail=` for shell, and every load goes through `chunkedTerminalWrite` in 32 KB slices. The route caps the body at `terminalBufferMaxBytes`, 32 MB by default, over a 100,000-line tmux history, so splitting with a long-lived shell tab pulls and parses that entire scrollback into a 5000-line xterm. Mirror the mode check and reuse a chunked writer.

## Smaller things, listed so you can fold them into the same push

- Pane B does not follow live appearance changes (`terminal-split.js:28`). `applyTerminalSkin()`, `applyTerminalFontFamily()`, `applyTerminalFontWeights()` and `setFontSize()` all propagate to the primary terminal, the echo overlays and teammate panes, but not to `this._splitPane.terminal`, so switching skin while split leaves Pane B on the old palette. It also hardcodes `fontSize: 14` and `scrollback: 5000` instead of reading `codeman-font-size`, `terminalFontFamily` and `DEFAULT_SCROLLBACK` (50000), which the spec says it should.
- The promotion path ignores `_closingSessions` (`terminal-split.js:375`). Closing Pane A's own tab makes the wrapper select Pane B while `closeSession()` is mid-await and about to select the next tab itself, so where you land depends on the race. That set exists for exactly this case: collapse the split, but skip the `selectSession` when `this._closingSessions.has(data.id)`.
- The desktop gate is a phone gate (`mobile.css:535`). The hide rule sits in the `max-width: 599px` block, so 600 to 1023px still gets the button while the divider only binds mouse events. `home-sessions.js` gates at 1180px in JS with a matching media-query backstop, which is what the spec points at.
- `.btn-split` is not in the `body.solo-mode` hide list (`styles.css:2556`), so a popped-out single-session window offers Split, right next to `.btn-multimonitor` which is hidden there.
- Detaching a session after the split is open re-creates the PTY-size fight the picker filter prevents at pick time (`terminal-split.js:169`). `detachSession()` (`app.js:1299`) does not know about the split, and `_sendResize()` has no `detachedSessions` yield.
- Docs: the invariants section is good, but CLAUDE.md gets only the load-order entry, no short rule and pointer, and line 174 still says "34 modules". The user-facing surfaces are `docs/wiki/The-Dashboard.md:118` (header button table) and `docs/wiki/Settings-Reference.md:58`.
- `terminal-split.js:8` points at `docs/superpowers/specs/2026-09-15-split-pane-sessions-design.md`, which is not in the repo; the doc is `docs/split-pane-sessions-plan.md`. The `@dependency` line on `:11` credits terminal-ui.js for `CodemanTerminalFont`, which lives in constants.js.
- `index.html:1928` reuses `data-preview-order="12"`, already taken by the Ultracode chip in the same `header` slot. It renders in the right order only because `Array#sort` is stable.
- `this._splitPane.connect()` at `terminal-split.js:291` is an un-awaited async call with no `.catch()`.
- The split row is 6px wider than its container (`styles.css:18594`): two 50% panes plus a `flex: 0 0 6px` divider, neither shrinkable, so `.main`'s overflow clips Pane B's right edge.
- Pane B's header and the picker rows render session names as plain text nodes (`terminal-split.js:277` and `:192`). `i18n.js` does exact-string lookup over text nodes and skips `.session-name` for that reason, so on zh-CN a session literally named "Sessions" gets translated.
- The Split button never changes class, title or aria state while a split is open, so nothing says which of its two behaviours a click will get.

One thing that is not a code change but belongs in the spec: while both panes take keyboard input, the global capture-phase shortcut handler (`app.js:1140`) always resolves against Pane A, so Ctrl+L or Ctrl+W typed into Pane B clears or closes the other session. Worth a line under the "deliberately plainer" heading.

Fix the blocker and the two majors and I will take another pass. Separately, I still owe you an answer on #452 about whether I want an in-page divider alongside the pop-out window, and I will get to that before you spend more evenings here. The picker, the prototype patches at module-evaluation time, the auto-collapse ordering test and the `--hidden` CSS static guard are all genuinely good work.

timkjr:
Thanks for catching that fast — pushed a fix for the blocker and both majors, plus the smaller list.

**Blocker.** You're right, moving the key into `displayKeys` was only half of it. Added `showSplitButton: _ssp,` to the strip destructure right next to `showFileViewerButton`, dropped the stray mention from the schemas.ts comment (a mention there reads as "this is a real field" to the next person grepping it), and added a guard test mirroring `test/terminal-auto-copy.test.ts`'s three-way rule so this can't recur silently.

**Divider drag.** Split `SplitTerminalPane.fit()` into `localFit()` (reflow only) and `fit()` (reflow + send). The drag now coalesces `onMove` into one `localFit()` per animation frame via `requestAnimationFrame`, and sends the real resize for both panes exactly once, at `onUp` — matching Pane A's own `throttledResize` convention.

**Scrollback.** Mirrored the mode check from `app.js`'s `selectSession()`: shell sessions get a bounded `?tail=` fetch instead of `?full=1`. Didn't reuse `chunkedTerminalWrite` directly — it's wired into session-switch generation counters and a live-output gate this simpler, independently created/destroyed pane has no equivalent of — so I wrote a minimal chunked writer scoped to `SplitTerminalPane` (32KB slices, yields a frame between each).

**The smaller list**, all addressed:
- Pane B now follows live skin/font-family/font-weight/font-size changes (propagates in all four functions, matching the `teammateTerminals` pattern), and reads real settings at construction instead of hardcoding `fontSize: 14` / `scrollback: 5000`.
- Promotion now skips `selectSession()` when `_closingSessions` already owns the delete.
- Detaching mid-split now yields the PTY size in `_sendResize()` too, not just at pick time.
- `.btn-split` joins the `body.solo-mode` hide list.
- Fixed the 6px overflow (`flex-shrink: 1` on both panes).
- Pane B's header and picker rows are `data-i18n-skip`/`.session-name` now.
- The Split button reflects open/closed state (`.split-open`, `aria-pressed`, a title that says what the next click does).
- `connect()` has a `.catch()`.
- Fixed the stale doc path and the wrong `@dependency` credit in the fileoverview.
- Fixed the `data-preview-order="12"` collision with the Ultracode chip (moved to 15).
- Finished the desktop gate properly this time: `SPLIT_PANE_MIN_WIDTH` (1180, matching `HOME_SESSIONS_MIN_WIDTH`) backs an actual JS check with a `matchMedia` listener, CSS backstop widened from 599px to 1179px so it actually covers the tablet range you flagged.

**Docs.** Added the short CLAUDE.md paragraph (was previously only in the load-order list), fixed the stale "34 modules" count, added Split to both wiki pages, and recorded the shortcut-handler note (Ctrl+L/Ctrl+W typed into Pane B affects Pane A) in the design spec under "deliberately plainer."

Full gate's green (7736 passed), plus the 7 split-pane browser tests, typecheck/lint/format/public-assets.

Ark0N:
Thanks for turning this around again, and for the care in the write-up. Everything from my last two passes checks out here: the `showSplitButton` strip out of the settings PUT with its own three-way guard test, the rAF-throttled drag with a single resize at drag end, the shell-mode `?tail=` fetch and the chunked writer, live appearance propagation, the `_closingSessions` guard on promotion, the 1180px gate, the solo-mode hide, the `detachedSessions` yield, the docs and both wiki pages. Full gate green here (409 files, 7736 tests, 563s), all three browser files pass, and CI is green.

One major left, then this can land.

## Major: app chords write control bytes into Pane B's live session

`src/web/public/terminal-split.js:82`

`SplitTerminalPane.connect()` opens its xterm and never calls `attachCustomKeyEventHandler`, so Pane B has none of the gates the primary pane installs at `terminal-ui.js:318`. The capture-phase handler in `app.js:1140` calls `preventDefault()`, and that does not stop xterm, which is the whole reason those gates exist (see the COD-153 comment at `terminal-ui.js:344-353`).

I measured it in a real browser against a real PTY, by patching Pane B's `ws.send` and pressing each chord with Pane B focused:

| chord | app action | what Pane B's socket sent |
| --- | --- | --- |
| Ctrl+K | palette opens | `{"t":"i","d":"\u000b"}` (readline kill-line) |
| Ctrl+L | clears Pane A | `{"t":"i","d":"\f"}` |
| Alt+1 | switches tab | `{"t":"i","d":"\u001b1"}` |
| Ctrl+W | kills Pane A's session | `{"t":"i","d":"\u0017"}` |

So opening the palette while typing in Pane B truncates what you had typed there, and Alt+tab-switching walks an escape sequence through it. The spec (`docs/split-pane-sessions-plan.md:93-99`) names the other half of this, that shortcuts always resolve against Pane A, but not the byte half.

What I would like: a custom key handler on Pane B's terminal returning `false` for the same set, routed through the registry-aware predicates so a rebind or a disable still restores plain terminal behaviour: `shouldOpenCommandPaletteFromShortcut`, the `/^(Digit[1-9]|BracketLeft|BracketRight|KeyK)$/` Alt-code test, and `shouldToggleSessionSidebarFromShortcut`. Worth deciding on Ctrl+V in the same pass, since `terminal-ui.js:408` routes it to the image paste trap and Pane B is on xterm's default. A browser test that presses Ctrl+K into Pane B and asserts its socket sent nothing would keep it fixed.

## Smaller things, if you are pushing anyway

- Narrowing the window past the gate hides the button but leaves an open split on screen (`terminal-split.js:224`). The `matchMedia` listener is already there; `if (!wide) this.closeSplitPane?.();` inside it would do. Both panes have `min-width: 240px`, so a narrow window overflows the row and `.main` clips Pane B's edge.
- Split is still offered while a web tab is active (`styles.css:15342` hides the container, nothing gates the button, and `activeSessionId` survives `openWebview()`), so picking a session there opens Pane B's socket behind a hidden container with nothing to show. An early return in `openSplitPane()` on `this.activeWebviewId` covers it.
- Pane B ignores `{"t":"r"}` (`terminal-split.js:129`). The primary pane routes it to `_onSessionNeedsRefresh()` (`app.js:2990`) and the server emits it when an external-CLI session finishes its first TUI render, so Pane B can sit on stale content with no way to ask for a reload.
- The divider drag has no `button !== 0` guard, no `preventDefault()`, no pointer capture and no body-level cursor/selection lock (`terminal-split.js:495`). Dragging selects the text under the cursor, and a right-click on the divider arms the drag and leaves the document `mousemove` listener attached. `tab-rail-resize.js:255-278` plus `body.tab-rail-resizing` (`styles.css:672`) is the pattern here.
- Pane B's input frames carry no `cid`/`seq` (`terminal-split.js:87`). `ws-routes.ts:176-178` supports that on purpose, so this is fine as a choice, but it belongs in the "deliberately plainer" list in the invariants entry next to the overlay and the IME rather than being silent.
- The `?full=1` capture is fetched before `_sendResize()` runs, so it is always a frame built at the pre-split geometry, written into a terminal at half the width (`terminal-split.js:108`). Worth one manual look with a claude session in Pane B; a repaint should correct it, but the full-capture path is fussy about row alignment.
- Live writes go straight to `terminal.write()` per frame with no in-flight gate (`terminal-split.js:133`), unlike the primary pipeline.
- `data-preview-order="15"` (`index.html:1928`) puts Split last in the header preview, after Lifecycle Log, while the real button sits between Multi-monitor and Ultracode Agents. `Number()` parses `11.5` fine.
- None of the Split strings have i18n entries, and `_updateSplitButtonState()` writes English `title`/`aria-label` at runtime, which would overwrite a translated tooltip. Every sibling header button has a zh-CN entry.
- Pane B's `x` and the picker rows are plain elements with inline `onclick`, so neither is keyboard reachable.
- `terminal-split.js:110` prefixes the fetch with `CodemanBase.base` even though the global `fetch` wrapper already does it. Harmless (the helper is idempotent), just the only site doing both. The WebSocket URL below it does need it.
- `.gitignore`'s `.worktrees/` and the `vi` import in `test/routes/session-routes-workspace-hooks.test.ts:22` are still not split-pane. The import is a no-op under `globals: true`.

Fix the key handler and I will merge. The picker, the prototype patches at module-evaluation time, the promotion-ordering test and the `--hidden` CSS guard are all good work, and that last one I am keeping whatever else happens.

One more thing I owe you: I still have not answered #452 on whether I want the in-page divider alongside the pop-out window. You have put three passes into this now, so let me settle that before you spend another evening on it.

timkjr:
Fixed the key handler, plus everything on the "if you're pushing anyway" list.

**Key handler.** `SplitTerminalPane.connect()` now installs its own `attachCustomKeyEventHandler`, gating the same three chords the primary pane gates: the command palette (`shouldOpenCommandPaletteFromShortcut`), the Alt+1-9/[/] tab-nav code test, and the sidebar toggle (`shouldToggleSessionSidebarFromShortcut`) — all routed through the registry so a rebind/disable restores plain terminal behavior. Ctrl+V stays on xterm's own default paste; Pane B has no image-paste trap to route it to, so intercepting it would only break paste. Added a real-Chromium test that dispatches Ctrl+K/Alt+1/Alt+B at Pane B's own textarea and asserts its socket sent no `{t:'i'}` frame for any of them.

**The smaller list:**
- Narrowing past the 1180px gate now closes an open split instead of stranding it on screen.
- Split refuses while a web tab is active (`activeWebviewId`) — it used to open Pane B's socket behind a hidden container.
- Pane B now handles `{t:'r'}` (the server refresh frame) via a `_loadBuffer()` helper shared with `connect()`, instead of ignoring it.
- The divider drag is pointer events + `setPointerCapture` now (mirrors `tab-rail-resize.js`), with a `button !== 0` guard, `preventDefault()`, and a `body.split-pane-resizing` cursor/selection lock — plain mousedown was selecting the text under the cursor as it crossed both terminals.
- Pane B's close control and the picker rows are real `<button>`s now (keyboard-reachable), with the CSS chrome reset back to the original look.
- Dropped the redundant `CodemanBase.base` prefix on the buffer fetch — the global `fetch` wrapper already applies it.
- `data-preview-order` for the Split chip moved to `11.5`, matching its real position between Multi-monitor and Ultracode Agents; widened `test/app-settings-structure.test.ts`'s regex to allow the decimal since `Number()` parses it fine for the preview sort.
- Added zh-CN entries for the Split button strings and the empty-picker message.
- Dropped the stray `vi` import — you're right it's unrelated, and it's ambient under `globals: true` anyway.
- Recorded the fix and the deliberate no-`cid`/`seq` choice in the split-pane-sessions invariants entry.

Full gate's green here too (409 files, 7736 tests), plus all 8 split-pane browser tests including the new one.

Ark0N:
Thanks for another fast turnaround, and for the write-up that came with it. The key handler is in, all three chords are routed through the registry-aware predicates, and everything else from the last list checks out here: the narrow-window close, the `activeWebviewId` refusal, the `{t:'r'}` handling, the pointer-capture drag, the real `<button>`s, the `11.5` preview order, the zh-CN strings. Full gate green (409 files, 7736 tests, 564s), all 8 browser tests pass, CI is green.

Two things before this lands, both small.

## 1. A split against a session with no PTY is a dead pane that eats what you type

`src/web/public/terminal-split.js:436`

`openSplitPane()` goes straight to `new SplitTerminalPane(...)` + `connect()`. `selectSession()` (app.js:6244-6279) does one thing first that this skips: when `session.pid === null` it POSTs `/api/sessions/:id/shell` or `/api/sessions/:id/interactive` (no body, so it cannot clear a tripped PTY-exit breaker) to re-attach the PTY.

Without it, nothing is reading the tmux pane, so no `terminal` event ever fires and Pane B never updates; and `Session.write()` is `if (!this.ptyProcess) return false` (src/session.ts:3582), so every keystroke is dropped. Pane B sends no `seq`, so there is no ACK either way and nothing reports the loss. The socket is healthy, so your new `[Pane B disconnected]` banner does not fire.

`pid` goes null on every local PTY exit, not just on a fresh create (session.ts:2504, :3009, :4026), so the reachable cases are a CLI that exited, a crash-looped session whose breaker tripped, or an attach that failed at restore. The tab stays in the strip and in the picker in all of them.

I measured it: real server, real Chromium, two shell sessions, only the first started. Made the first active, called `app.openSplitPane(<unstarted id>)`. The split opened, `_wsReady` went true, I sent `{"t":"i","d":"echo hi\r"}` on Pane B's own socket, waited 2.5s, and Pane B's xterm buffer came back empty. A session whose tmux pane still exists is worse: `_loadBuffer()` returns the capture, so the pane looks populated and correct while being completely frozen.

Either mirror `selectSession()`'s attach in `openSplitPane()` before you build the pane (leaving `respawnBlocked` alone, or refusing with a toast), or exclude `pid === null` sessions from `buildSplitPickerSessions()` the way detached ones are excluded, and say so in the invariants entry. The picker exclusion is fine for v1 if you prefer the smaller change.

## 2. The chord test passes with the gate deleted

`test/split-pane-terminal.browser.test.ts:218`

The three synthetic keydowns carry no `keyCode`, and `keyCode` is what xterm's `evaluateKeyboardEvent` switches on to pick the bytes. At `keyCode: 0` nothing can produce a data frame, so the assertion holds no matter what the custom key handler does.

I deleted the whole `attachCustomKeyEventHandler` block from `terminal-split.js` and re-ran the file: 3 passed, this test included. Adding `keyCode: 75 / 49 / 66` makes it fail without the gate, which is what we want.

With the gate back and `keyCode` set it fails on exactly one frame, `{"t":"i","d":"\u001bb"}`, and that one is correct: `shouldToggleSessionSidebarFromShortcut()` is gated on `isSessionSidebarActive()` (app.js:4325) and the default layout is the header strip, so Pane A forwards ESC b there too. So: add the three `keyCode`s, and either drop the Alt+B case or switch the layout to sidebar before asserting on it.

The gate itself is right. Only its test is hollow.

## Smaller, take or leave

- `terminal-split.js:101` says the handler mirrors the primary pane's gates. It mirrors three of seven. The ones missing with real consequences: Ctrl+Z, which Pane A swallows for every non-shell mode on purpose (terminal-ui.js:412-432, suspending an unattended agent loop), and Shift+Enter, which in Pane B emits plain `\r` and submits the prompt instead of inserting a newline. Smart-copy Ctrl+C is the third. At minimum name them in the "deliberately plainer" list in the invariants entry next to the overlay and the IME; porting the Ctrl+Z swallow would be better, it is about six lines.
- `terminal-split.js:131`: `connect()` awaits `_loadBuffer()` before opening the socket, so output written in that window is in neither and is gone from Pane B's scrollback. Small, but the primary pane opens first and queues.
- `terminal-split.js:150`: the `{t:'r'}` path clears and then fires an unawaited `_loadBuffer()`, so live frames landing during the fetch get buried under the replay, and there is no cooldown on repeated refresh frames.
- `.gitignore:6`: `.worktrees/` is still on the branch.
- CLAUDE.md:359's z-index list did not get `.split-picker-menu` (1000).
- `docs/split-pane-sessions-plan.md:3` still reads `Status: Draft, pending review`.

I also owe you the answer on #452 that I said I would settle, and I have not. Do not spend another evening on this until I do. The two items above are small enough that I am happy to apply them at merge if that is easier for you; say the word.

Ark0N:
#452 is settled: yes, the in-page divider goes in alongside the pop-out window, not instead of it. The reasoning and the v2 boundaries are in the thread: https://github.com/Ark0N/Codeman/discussions/452#discussioncomment-18529622

So this is unblocked. The two items from my last pass (the `pid === null` split and the hollow chord test) are all that stand between it and a merge. Push them, or say the word and I apply them on the way in.

timkjr:
Pushed. Both items from your last pass are in: `buildSplitPickerSessions()` now excludes `pid === null` sessions (the picker-exclusion option you called fine for v1), and the chord test carries real `keyCode`s now — `keyCode: 75/49/66` plus the three you flagged, all asserted against the actual gate rather than a synthetic event xterm would ignore regardless.

One more went in that you didn't ask for: Ctrl+Shift+C. Looking at the "smaller, take or leave" list again, the copy gate you'd already flagged as missing three chords (Ctrl+Z, Shift+Enter, smart-copy) turned out to be four — the block only entered its copy logic behind `hasSelection()`, so a selection-less Ctrl+Shift+C fell through to `return true` with no `preventDefault()`, same shape as the other three. I checked what actually reaches the PTY there before assuming the worst: xterm's `evaluateKeyboardEvent` never emits data for a shifted ctrl-letter at all, gate or not, so it was never an interrupt risk — just Pane B silently handing the chord to the browser instead of honoring the "never falls through" contract Pane A gives it. Ported the same `ev.shiftKey` branch, with a test that asserts on `event.defaultPrevented` rather than the frame count, since the frame count passes either way for this key.

#452: thanks for the go-ahead, and noted for v2: persistence, drag-to-divider, and echo-overlay parity last since it's mobile.

Full gate green (409 files, 7738 tests), typecheck/lint/format clean, both browser suites for this file passing including the new one.

Ark0N:
Merged into master for 1.32.0, with the two minors applied on the way in (299a21d5): a split that collapses mid-drag now releases the drag (class, pointer capture, listeners, queued frame), and openSplitPane() re-applies the picker's exclusions before touching the DOM. The nits went in too: onclose nulled in destroy(), the unread data attribute dropped, no resize aimed at a session the server just removed, and the refresh path is single-flight. Thank you for five rounds in two days, and for measuring the pointer-capture case in a real browser.

PR #454, opened 2026-09-19
fix(run): make the Instance count stepper work for every non-Claude mode

timkjr:
## Summary
- The "Instance count" stepper next to the Run button was only ever read by `runClaude()`. Every other mode's run function — OpenCode, Codex, Gemini, Antigravity, Pi, OMP, Grok, DeepSeek — hardcoded a single quick-start call, so bumping the counter while on any of those modes silently launched exactly one session with no error.
- Extracted the shared launch-N-and-select-the-first loop into `_launchQuickStartInstances()`, reused by all eight modes, plus `_readTabCount()` for the shared clamp/parse. Each mode still builds its own quick-start body (config differs per CLI) via a closure passed into the shared loop.

## Test plan
- [x] `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run check:frontend-syntax`
- [x] `npm test` (full suite, 7368 passed) — updated `test/run-mode-ui.test.ts` fixtures for the five modes with existing coverage (Codex, Gemini, Antigravity, Pi, Grok) to include the stepper element
- [x] Manually verified live across Pi, OMP, and Claude with counts of 2–3

https://claude.ai/code/session_012AcpEgRwdEbdTqm4W4wanS


Ark0N:
This is going in. Thanks for this, and welcome.

You found a real silent-failure bug, which is the best kind of first contribution: the stepper sat there next to the Run button looking like it worked, and on eight of the ten run modes it did nothing at all. No error, no hint, just the wrong number of sessions. The refactor is also the right shape. Each mode keeps its own quick-start body through the `buildBody` closure, so no payload changed (I diffed all eight against their pre-PR form), and you kept the two things that are easy to lose here: the `{success,data}` envelope unwrap, and `_ensureCreatedSessionVisible()` before `selectSession()`.

What I am folding in on top at merge:

**A regression test.** This is the one thing the PR was missing. Your five test edits are fixture maintenance for the single-instance case, so they pass identically with and without the change, and three of the eight converted modes have no quick-start test at all. The new case drives `runGrok()` with `tabCount` at 3 and asserts three POSTs with `w1`/`w2`/`w3` names and the first session selected. I checked it fails against master's `session-ui.js` (`['w1-grok-case']` vs the expected three) and passes on yours, so it genuinely pins the behaviour rather than just describing it. This matters more than box-ticking here, because the bug happened in the first place by each mode carrying its own copy of the launch call.

**The count in the opening banner.** Each caller announced `Starting Grok session in ...`, singular, and then the helper printed `Starting 3 Grok session(s) in ...` right after it, so a home-screen launch got two headers in a row. Worse, with another session already active `_appendSessionLaunchStatus` returns early, so the helper's lines were suppressed entirely and a five-instance launch showed one toast promising one session. The callers now read the count first and put it in their own banner, the way `runClaude()` already did, and the helper's opener is gone.

**`runClaude()` now calls `_readTabCount()`** instead of keeping its own copy of the 1..20 clamp, which is what your JSDoc was already pointing at.

**A `?.` on the element read**, because hoisting the count above each caller's `try` block means a missing `#tabCount` would otherwise throw somewhere the launch-error path cannot report it.

Two things I did not change, on purpose. The helper launches sequentially where `runClaude()`'s local path fans out, but that matches `runClaude()`'s own remote loop and is the safer default, so I left it. And a partial failure leaving the started sessions unselected behaves exactly like the remote loop already does, so fixing it is a consistency change for both, not for this PR.

One genuinely pre-existing thing you surfaced: the welcome-screen buttons call `app.runOpenCode()` and friends directly rather than going through `app.run()`, so they skip the in-flight lock, and now that the other eight modes honour the stepper a double click there produces N duplicate names instead of one. That predates you. I will take it separately.

On #453, which is yours too: it is not being ignored. It is waiting on Discussion #452, because whether Codeman wants in-window split panes at all is a product call I owe you before you spend more time on a 3000-line PR. I will answer there.

PR #460, opened 2026-09-20
feat(install): three questions up front, an unattended build, and a URL you can scan

Ark0N:
## Why

`curl -fsSL https://getcodeman.com/install | bash` asked roughly ten questions on a fresh box, half of them after a multi-minute build, with several sudo password prompts scattered between them, and the question that matters most (how do I reach the dashboard) came last. Nobody could walk away while it ran. The Tailscale route worked but ended on the machine's tailnet name with no way to influence it, gave up when another app already owned `:443`, and once the terminal scrolled away nothing told you the URL again.

## What changes for someone running the one-liner

The installer now looks first, asks everything up front, works unattended, and ends on the URL.

```
  System      linux (x64), debian
  Found       git 2.43.0, tmux 3.4
  Missing     Node.js 22, build tools (make, g++)
  AI CLIs     none found
  Tailscale   not installed
  Existing    none

Install the missing pieces now (Node.js 22, build tools (make, g++))? [Y/n]
==> Some steps need administrator rights. You will be asked for your password once.

  1/3  How should the Codeman dashboard be reachable?      (Tailscale / your network / this machine)
  2/3  Name this machine on your tailnet?                   (Tailscale route only, default no)
  3/3  Run Codeman in the background?                       (Enter installs the service)

==> Cloning Codeman (4s)
==> Installing dependencies (7s)
==> Building Codeman (24s)
==> Systemd service installed and started

  Codeman 1.31.0 is running
    Your tailnet    https://<node>.<tailnet>.ts.net   (HTTPS, any of your devices)
    This machine    http://localhost:3000
    [QR code]       scan to open it on your phone
    Manage    systemctl --user restart codeman-web   journalctl --user -u codeman-web -f
```

- **Every step that needs a human runs before the build**: one consent for all missing packages, one sudo prompt kept warm for the run, the AI CLI menu, and the whole Tailscale preflight (install, the login URL, the operator grant, the tailnet HTTPS toggle, which is now polled with the admin page opened in a browser instead of a "re-check now?" loop).
- **The work runs behind spinners** with output in `~/.codeman/install.log`; the tail is shown on failure, and a failed dependency install names its step instead of exiting silently.
- **The done screen** leads with the URL and a terminal QR code (the `qrcode` package Codeman already depends on). `install.sh status` prints it again any time.
- **Flags** pipe through `bash -s --`: `--tailscale | --lan | --local`, `--name <n> | --no-rename`, `--service | --run | --no-start`, `--yes`, `--password`, `--port`. Env-var twins are unchanged. `--port` is now also written into the service file, which it never was.
- **Service is the Enter default** on question 3 (it used to re-prompt with no default). The cloudflared question leaves the main flow (`install.sh cloudflared`).
- **"Continue WITHOUT a password?" defaults to yes** now (maintainer decision; the loud warning and the red closing notice stay).

## The Tailscale part

Researched against Tailscale's current docs and the 1.102 CLI:

| You want | How | In this PR |
| --- | --- | --- |
| The machine's existing name (`https://tnode.<tailnet>.ts.net`) | nothing | the default, unchanged |
| `https://codeman-<hostname>.<tailnet>.ts.net` | `tailscale set --hostname` | opt-in question 2, `--name`, `install.sh name`; **defaults to no everywhere** because the tailnet name is also the machine's SSH identity; uninstall offers the old name back |
| `https://codeman.<tailnet>.ts.net` | a Tailscale Service | docs hint only: needs a tagged node (a user-signed-in device cannot host one) plus admin-console definition and approval, still beta |

The flow is now two halves on purpose: `tailscale_prepare` (question phase) decides the serve *shape*, `tailscale_apply` (after the build) issues the one serve command. Deciding the shape early matters because it can change the service unit: **when `:443` already belongs to another app** the default is now a sub-path, `https://<node>/codeman` (`tailscale serve --set-path /codeman` plus `CODEMAN_BASE_URL=/codeman` in the unit), else a second port, replace, or skip. Measured live: serve strips the mount prefix before proxying, Codeman's ingress tolerates unprefixed requests, and `--base-url` is what makes the emitted URLs carry the prefix. Serve config is keyed by the DNS name it was written under, so a rename takes our mapping down first and re-adds it under the new name (the only other cleanup would be `serve reset`, which stays forbidden).

Safety rules pinned by `test/install-sh-invariants.test.ts`: never `serve reset`, never `funnel`, never a Tailscale Service, every serve mutation through the sudo-aware wrapper, rename before shape, every flag documented in the header, rename default no, and `npm install` with `CODEMAN_NO_AUTOSTART=1` (below).

## A pre-existing bug this surfaced

`npm install` runs `scripts/postinstall.js`, which builds and starts a **detached** `codeman web` on `127.0.0.1:3000` unless `CODEMAN_NO_AUTOSTART` is set. The installer never set it. On a fresh machine the service written a minute later crash-looped on `EADDRINUSE` while the done screen reported "running" off the orphan, and the dependency step built twice. The installer now sets the opt-out (dependency step 58 s to 7 s in the sandbox).

Two more: the LAN address now comes from the default route's source address instead of the first interface `hostname -I` lists (a docker or LXD bridge, or tailscale, can be first), and a foreign `/Library/LaunchDaemons/com.codeman.web.plist` (a headless Mac's own supervisor) is left alone instead of being replaced by a LaunchAgent.

## Verified

- Walked through interactively, prompt by prompt, in a fresh Ubuntu 24.04 LXD container (systemd, non-root user with a sudo password, no Node, no build tools): fresh install on the network route with the service, a re-run with `--lan --service`, `install.sh status`, `install.sh uninstall`. Exactly one server, owned by the unit, on `0.0.0.0:3000`; uninstall left nothing.
- Sub-path mode end to end over a real tailnet: an isolated Codeman on port 3999 with `--base-url /codeman` behind `tailscale serve --https=8445 --set-path /codeman` answered the page (with `<base href="/codeman/">`), the hashed assets, the no-slash path and the SSE stream.
- A non-interactive `CODEMAN_TAILSCALE=1` install into a sandboxed HOME on a node that already serves Codeman took the no-mutation keep path and rendered the QR on a pty.
- Every decision function driven with stubbed tailscale state under bash 5.2 and the `bash:3.2` container CI uses (flags, launch default, the four answers for an occupied `:443`, the three serve commands, the rename question, `run_step`, the unit round-trip of base URL / port / an escaped password, the HTTPS-toggle poll). The CI bash 3.2 step now runs that drive too.
- `bash -n`, the invariants and detection-parity tests, and `cli-catalog-sync` pass.

## Not yet exercised on a fresh machine

Tailscale absent, logged out, or with the HTTPS toggle off; a real rename against a live node (the off-rename-re-add order is unit-driven only); macOS; uninstall of a serve mapping. `docs/installer-v2-plan.md` section 8 lists the matrix and the venues.

## Docs

README, wiki `Installation`, `Remote-Access` (naming table, the occupied-`:443` options), `Running-As-A-Service` (macOS login caveat, LaunchDaemon), `docs/security-architecture.md`, CLAUDE.md. `docs/installer-v2-plan.md` is the design plus the verification record; the earlier `docs/tailscale-installer-plan.md` points at it.

## Release

No changeset in this PR; suggested bump: minor (new flags, new subcommands, new behaviour on a fresh install). The getcodeman.com copy still describes the old flow and lives outside this repo.

## Review fixes (2026-09-21)

Both reviews on this PR (the DeepSeek Harness pass and the Claude pass) are folded in, each pinned in `test/install-sh-invariants.test.ts` and, where it is bash logic, driven in the `bash:3.2` CI step too:

- **Before merge, from the reviews:** the done screen's Start line is composed from every non-default value in one place (`start_command_hint`, shared with the `exec` branch as `export_bind_env`), so "do not start" under a sub-path or a custom port no longer prints a bare `codeman web`; the `--lan` / `--tailscale` / env preset paths keep an existing password instead of rewriting the unit open with the unauthenticated ack; `--password` and `--port` flip `RECONFIGURE` so they reach the unit instead of taking the quiet update path; `install.sh name` re-syncs the unit's base URL after a rename re-adds the mapping.
- **Merge-time items:** the sudo keepalive is ended before the `exec` into the foreground server (exec skips the EXIT trap); Ctrl+C in the HTTPS-toggle poll skips Tailscale for the run instead of killing the installer (INT trapped for the poll only, verified by signalling a live poll); `uninstall` asks before removing a LaunchDaemon it never wrote; a foreign LaunchDaemon gets a `launchctl kickstart` hint and the done screen stops claiming the new build is running; the preflight summary reads the Tailscale state without node (line grep on `tailscale status --json`); the LAN security notice uses the configured port; a bare re-run ends on the done screen; a build failure after a rename names the `install.sh tailscale` recovery; `TS_JOINED_HERE` is gone; the plan doc and `docs/architecture-invariants.md` say what the code does.
- A `minor` changeset is included.


Ark0N:
_Reviewed by qwen3.8-27b at medium effort on a local RTX 5090, through the DeepSeek Harness (dsh)._

Three questions up front, walk away during the build, and end on the URL with a QR code. Everything checked out (invariants, parity, catalogue sync, the bash 3.2 drive, and a local repro of the CI question-phase run), and the CODEMAN_NO_AUTOSTART fix for the postinstall orphan is well documented and pinned by a test.

One thing to fix before merge:

- install.sh:2988: tailscale_apply runs regardless of the launch choice, so with Tailscale chosen, :443 occupied by another app (path mode) and both service questions answered no, the installer creates the https://<node>/codeman mapping but nothing ever runs with CODEMAN_BASE_URL=/codeman, and the done screen tells the user to start with plain codeman web. That leaves a URL the user cannot open, with the mapping persisting across reboots. Print the start command with the base URL when LAUNCH_CHOICE=3 and BIND_BASE_URL is non-empty (and the same for setup_name_subcommand, which re-adds a path-mode mapping after a rename without calling sync_service_base_url).

Small things I would fold in at merge time:

- install.sh:2067: TS_JOINED_HERE is set but never read now that the rename defaults to no everywhere; drop it or keep it with a reason.
- docs/installer-v2-plan.md:91 and :355 still propose the rejected rename default (YES when this run joined the tailnet); mark them decided so the header and the body agree.
- docs/architecture-invariants.md:13: the installer note still says the network bind gets a required password prompt; the skip confirmation now defaults to yes, so the note is a line out of date.

Nothing else to change. I will take these in on merge and ship it as a minor bump, since the new flags and subcommands are user-facing.

Ark0N:
_Second review, by Claude Fable 5.1, after the DeepSeek Harness pass above._

I re-ran the three vitest files (55 pass), the bash 3.2 question-phase drive from ci.yml in the `bash:3.2` container (passes), and CI is green on this head. I also drove the rename dance against a fake `tailscale` binary for both shapes, and the argv order is exactly what the plan asks for: `serve --https=443 off`, then `set --hostname`, then `serve --bg 3999` (and `serve --https=443 --set-path /codeman off` / `serve --bg --set-path /codeman 3999` for the path shape), with the rename record written. The `${x:+$x$'\n'}` join in the unit writers works under bash 3.2 too (checked in the container), the grep for bash 4 constructs comes back empty, and `CODEMAN_NO_AUTOSTART=1` really does skip both the build and the start in `scripts/postinstall.js`.

The blocker in the review above reproduces: with "do not start" and path mode the done screen prints `Start  codeman web` under a `This machine  http://localhost:3000/codeman` line. Its three small things hold too (`TS_JOINED_HERE` is written at install.sh:2067 and read nowhere; the plan still says "default YES" at lines 91, 328 and 355; the installer note in architecture-invariants.md:13 still says LAN with a required password prompt).

Two more to fix before merge:

- install.sh:1573 and :1589: **a flag re-run drops an existing password.** Both preset branches of `choose_network_binding` set `BIND_PASSWORD="${CODEMAN_PASSWORD:-}"` and never look at `EXISTING_PASSWORD`, even though `read_existing_binding` already ran in `preflight_detect`. So `bash -s -- --lan --service` on an install whose unit carries a password rewrites the unit without it AND with `CODEMAN_ALLOW_UNAUTHENTICATED_NETWORK=1`, with no "Continue WITHOUT a password?" question, and `--tailscale --service` drops the password the same way. Driven through the sourced script: `--lan` gives `BIND_PASSWORD=[] BIND_ACK=1`, `--tailscale` gives `BIND_PASSWORD=[]`. The interactive path keeps the password on purpose ("never silently loosen"), and the env twins had this hole before, but the flags are the documented one-liner now. `${CODEMAN_PASSWORD:-$EXISTING_PASSWORD}` in both branches, and derive `BIND_ACK` from the result.
- install.sh:3359-3365: **the Start line drops the port as well as the base URL.** With `--port 4001` and "do not start", the two URL lines say `:4001` and the Start line says `CODEMAN_HOST=0.0.0.0 codeman web`, which binds 3000. Same fix as the blocker above: build that line from every non-default value (host, password, ack, base URL, port) in one place. The `exec` branch at :3016-3023 already exports exactly that set, so the two can share it.

Things I would fold in at merge time:

- install.sh:3023: **the sudo keepalive outlives the exec.** `exec node` replaces the shell without running the EXIT trap (checked: `trap ... EXIT; exec sleep 2` runs no trap), and the keepalive loop keys on `$$`, which is now the server's pid. So with "start in this terminal" the sudo timestamp is refreshed every 50 s for as long as the foreground server runs, instead of lapsing after sudo's 5 minutes. Kill `SUDO_KEEPALIVE_PID` (and the spinner, for the same reason) right before the `exec`.
- install.sh:2138: **Ctrl+C in the HTTPS-toggle poll aborts the whole run.** The text says "Ctrl+C to give up" and the plan says it falls back exactly as before, but the only trap is `EXIT`, so Ctrl+C ends the installer before the build and the user re-answers everything on the retry. Either trap INT for the length of the poll (return 1, which lands on the retrofit hint) or say that Ctrl+C aborts.
- install.sh:3492: **`uninstall` still `sudo rm`s a foreign LaunchDaemon.** The install side now declares that this installer never writes one, so any daemon there is the user's; uninstall then deletes it without a question. On the Mac mini that is the hand-written daemon. Ask first, or leave it and print the unload command.
- install.sh:3627 and :3632: **`--password` and `--port` alone do not reconfigure.** Neither flips `RECONFIGURE`, so on a completed install `bash -s -- --password x` takes the quiet update path and the password never reaches the unit, and `--port 4001` leaves the unit on 3000 while every URL hint, the serve target and the repair offer reason about 4001. Set `RECONFIGURE` for both.
- install.sh:2660 and :3346: **with a foreign LaunchDaemon the done screen says the new version "is running"** while the daemon still runs the old build, and nothing restarts a daemon (neither `main` nor `update`). One line under Manage: `sudo launchctl kickstart -k system/com.codeman.web`.
- install.sh:3048: **the preflight summary needs node to see a connected Tailscale.** `preflight_detect` only reaches "connected" when `command -v node` succeeds, so the very first screen on a fresh box that is already logged in (the box this PR is for) says `installed, not logged in`. A grep for `"BackendState": "Running"` on `tailscale status --json` settles it without node.
- install.sh:259: the LAN-with-password security notice still says "port 3000"; `--port` exists now.
- install.sh:3456: **a bare re-run ends on the security notice, not the done screen.** The re-run is how everyone updates, and the first thing a user tries when they want the URL back. `print_done_screen "" ""` there too.
- A build that fails after a rename took our mapping down (the rename runs in the question phase) leaves nothing fronting Codeman and the failure advice does not say so; `install.sh tailscale` recovers, so one line in `cleanup` when `TS_MAPPING_REMOVED_BY_RENAME=1`.

Same verdict as above: merge with these folded in, minor bump. The password drop is the one I would not ship without.

Ark0N:
Both reviews are folded in as one commit on this branch: 72d437ab.

- The two pre-merge items: the Start line now comes from `start_command_hint` (the same five values the `exec` branch exports through `export_bind_env`), so "do not start" prints `CODEMAN_BASE_URL=/codeman codeman web` under the path URL and `CODEMAN_HOST=0.0.0.0 CODEMAN_PORT=4001 codeman web` under a custom port. The `--lan` / `--tailscale` / env preset paths read `${CODEMAN_PASSWORD:-$EXISTING_PASSWORD}`, so a flag re-run keeps the password the unit already carries and never writes the unauthenticated ack over it.
- The merge-time list: `--password` and `--port` flip `RECONFIGURE`; `install.sh name` re-syncs the unit after the mapping is re-added; `stop_background_helpers` runs before the `exec`; Ctrl+C in the HTTPS-toggle poll is trapped for the poll only and returns to the retrofit hint (signalled a live poll to check, and the default INT disposition comes back afterwards); `uninstall` asks before touching a LaunchDaemon it never wrote; a foreign daemon gets the `launchctl kickstart -k` hint and the done screen says "is built" rather than "is running" for it; the preflight reads `tailscale status --json` with a line grep when node is not there yet; the security notice uses the configured port; a bare re-run ends on the done screen; a failed build after a rename names `install.sh tailscale`; `TS_JOINED_HERE` is gone; the plan doc and architecture-invariants now match the code.
- Pinned: nine new cases in `test/install-sh-invariants.test.ts` (static guards plus a real-bash drive of the kept password and the Start line), and the `bash:3.2` CI step drives the kept password, the Start line and the `--port` reconfigure too. A `minor` changeset is in the PR.

Holding the merge for now.

Issue #464, opened 2026-09-21
Text gets muffled sometimes, in both TUI default and fullscreen


rounakdatta:
<img width="630" height="1375" alt="Image" src="https://github.com/user-attachments/assets/aa86293d-b30d-4226-8d33-40900a23e3bf" />

This muffling of text is my biggest problem in using this software on phone. I'm unable to fully reproducce why this is happening, and how we can have a reliable fix for it.

Ark0N:
Hey thanks for the report, can you give me more information about which setup you are using? Which browser and version, which android, how do you connect?

rounakdatta:
I'm using Android 16 (OriginOS 6) and this is Chrome browser. My setup is a bit complex - https://github.com/rounakdatta/agentfest - this is my packaging over vanilla codeman. I have a full fledged deployment of this on a VPS and a neat browser URL (not plain Tailscale based) via which I connect.

rounakdatta:
Found it. Fix is in #431.

**It is not a rendering glitch — it is the browser terminal and the PTY disagreeing about how wide the screen is.**

Claude Code's TUI wraps its frame at the width the PTY reported, and erases the previous frame by moving the cursor up the number of rows it *believes* that frame occupied. When the browser terminal is a different width, each line takes more physical rows than the CLI counted, so its erase clears too few of them and the new frame paints over rows that were never cleared. That is exactly the two things in the screenshot: lines appearing twice, and a short tool summary sitting inside a longer sentence with the tail of the sentence still showing.

Reproduced against the same xterm this app ships — a PTY that thinks it has 120 columns, rendered into a 62-column terminal, draws every wrapped line twice.

### Why it was so hard to pin down

Because nothing could see it. Resize was **write-only**: the browser told the server a size and never learned whether it took. Four separate things could make the two disagree, all silently:

- `fitAddon.fit()` sized the terminal to the raw measurement, while the server was told that measurement floored at 40x10. Any time the floor applied, they diverged. **Raising the font on a phone is enough** — at font size 44 on a 430px screen, the measurement is 13 columns, the server was told 40, and the terminal stayed at 13.
- Two paths deliberately withheld the resize from the server (while the keyboard is up, and for a session popped out into its own window) but still reflowed locally — which is the one combination that cannot be right.
- Changing the font size, family or weight moves the cell size, so it changes the column count. None of them told the server anything.
[...]

PR #470, opened 2026-09-22
fix(terminal): recover a dropped output frame, do not merely schedule it

rounakdatta:
> **Stacked on #431.** Until that merges, the commit list below carries its four geometry commits too — the reviewable surface here is the single commit `00f022cc`, and the diff collapses to just that once #431 lands. Happy to rebase onto master the moment it does.

This is the follow-up I offered at the end of the #431 review, kept separate because that PR was already carrying six concerns and you have twice asked for smaller ones.

## The defect

`_onSessionTerminal` drops an incoming frame once the app-owned render queues already hold 128KB (`app.js`). That is the right call — the alternative is an unbounded backlog — but a hole in a TUI byte stream is a desynced cursor, and a desynced cursor is muffled text. The drop was only half of it; the recovery has to actually happen.

It was a fire-and-forget timer:

```js
if (!this._clientDropRecoveryTimer) {
  this._clientDropRecoveryTimer = setTimeout(() => {
    this._clientDropRecoveryTimer = null;   // handle gone first
    this._onSessionNeedsRefresh();          // skipped? nobody finds out
  }, 2000);
}
```

and `_onSessionNeedsRefresh` opens with four early returns, two of which are:

```js
if (this._isLoadingBuffer) return;
if (this._terminalRefreshOwner?.sessionId === sessionId) return;
```

**Those are most likely to be true during a heavy output burst — which is exactly when a drop happens.** A buffer load in flight, or a refresh already owning the session, and the recovery is skipped with its timer already cleared. The dropped bytes are never replayed, and nothing ever tries again.

It is the same defect you caught in your second round on `_wsOutputGapSession` — *"clearing only after a repaint actually happened would close it"* — sitting in the one recovery path where that principle had not been applied.

## The fix

`_onSessionNeedsRefresh` returns whether it actually repainted (true only where it settles `_markTerminalBufferReconciled`, false from every early return and the catch). `_scheduleDroppedOutputRecovery` re-arms while it has not.

Bounded by `DROP_RECOVERY_MAX_ATTEMPTS`: every reason the refresh can be skipped is transient contention that clears in seconds, and a permanently failing refresh must not become a loop against the API. Giving up at the cap leaves exactly what the old code left, so the floor is no worse than before. The same 2s debounce still collapses a burst of drops into one attempt — twenty dropped frames are still one fetch.

Callers that ignore the return value are unaffected; all five of them already did.

## Testing

Twelve cases in `test/dropped-output-recovery.test.ts`: the retry policy as a pure function in `constants.js` so the gate can reach it, the real scheduler from `app.js` under a fake clock, and source guards that the drop path goes through the scheduler and that no early return leaves a bare `return`.

⚠️ **The retry case and the no-retry case only pin this as a pair.** Either alone passes against something wrong — one against the old fire-and-forget timer, the other against retrying forever. I checked that rather than assuming it: reverting `app.js` to the old shape fails three of the twelve, and restoring it passes all twelve.

Two harness details that would otherwise have made the tests lie, both worth stating because the second is a real bug beyond the test:

- The vm context baked in the real `setTimeout`, so `vi.useFakeTimers()` could not reach the scheduler and every case reported zero calls. It delegates lazily now.
- `app.js` reached `CodemanDroppedOutput` as a bare global — fine in a browser, undefined in the vm harnesses the gate runs `app.js` under. That call sits **inside a timer**, where a `ReferenceError` is swallowed and would have taken the recovery with it. It reads through `window.` now, like `terminal-ui.js` does with its own constants.

**Gate**: 420 files, 7910 tests — 7897 passed, 12 skipped, 1 failed. The `cron-service` `/etc/passwd` failure is my box: it fails identically on a clean `origin/master` worktree and passes on CI. typecheck, lint, format and frontend-syntax are clean.

CLAUDE.md carries the invariant beside the render cap it belongs to.

## Scope

Deliberately only this. Two other paths I looked at and left alone, because both are **capture-only** — `stripInkRedrawBloat` is called only from the terminal route and `formatPaneSnapshot` only from `capturePaneBuffer`, while the live path is `emit('terminal', data)` straight to the socket. Neither can explain corruption in live output, which is what #464's screenshot shows. Happy to look at either if you want them, but they are not this.


Ark0N:
Merged together with #431, thanks @rounakdatta! This ships in 1.32.1.

Reporting whether the refresh actually repainted, and re-arming until it did, is the same principle #431 settled for the WebSocket gap marker, applied to the one path still trusting a timer. Keeping it off #431 made both easier to read.

Two fixes at merge, both from the review:

- The `TERMINAL DROP` crash-trail line sat on the drop path itself, so a burst could write it ~125 times a second and push everything else out of the 50-entry trail. It now fires once per debounce window, carrying the queued byte count.
- A refresh that died at the fetch deadline was retried four more times, and for a `?full=1` capture that deadline can be 45 to 120 s. `_onSessionNeedsRefresh` now returns `'deadline'` in that case and the scheduler does not retry it; the genuinely transient skips still retry as you wrote them. The CLAUDE.md sentence claiming every skip reason is transient contention is corrected.

